PR #5, opened 2023-12-17
chore: generate journal gql types

thevaibhav-dixit:


cala-server/schema.graphql
```diff
@@ -51,8 +51,25 @@ enum DebitOrCredit {
...
+type Journal {
+	id: ID!
+	journalId: UUID!
+	name: String!
+	externalId: String
+	status: Status!
+	description: String
+}
+
+input JournalInput {
```
bodymindarts, line 63: Should be CreateJournalInput

cala-server/schema.graphql
```diff
@@ -51,8 +51,25 @@ enum DebitOrCredit {
...
+	id: UUID
+	name: String!
+	externalId: String
+	status: Status!
```
bodymindarts, line 67: should have a default

PR #17, opened 2024-05-02
test: bats boilerplate

bodymindarts:


docker-compose.override.yml
thevaibhav-dixit, line 1: why are we deleting this file. I thought this contains the overrides for the otel receiver/

bodymindarts: This hasn't been needed for a long time. We used to need override files before we enforced a single test job running on our docker / nix hosts. Now that only 1 rtest runs at a time there is no port conflict with other projects so we can delete those files.

PR #42, opened 2024-05-10
test: improve bats idempotency

vindard:


examples/rust/src/main.rs
```diff
@@ -34,10 +35,12 @@ async fn main() -> anyhow::Result<()> {
         .outbox(OutboxServerConfig::default())
         .build()?;
     let cala = CalaLedger::init(cala_config).await?;
+    let random_number = rand::thread_rng().gen_range(0..1000);
```
bodymindarts, line 38: For this I had the idea that we could perhaps inject a test prefix:


${EXAMPLE_PREFIX}. Don't know if you think it would help - but we can set it to something random in the test as well as default it to something random in the example/src/main.rs if its not set.

vindard: Env variable idea makes sense. Would we ever want to move to having any sort of arg parsing for these?

bodymindarts: interesting question. I don't think so - certainly most of the time we would want them to execute with defaults.

bats/examples.bats
```diff
@@ -23,18 +32,18 @@ teardown_file() {
       }
     }'
   )
-
   exec_graphql 'cala-outbox-import-job-create' "$variables"
-
   name=$(graphql_output '.data.calaOutboxImportJobCreate.job.name')
-  [[ "$name" == "rust-example" ]] || exit 1;
+  error_msg=$(graphql_output '.errors[0].message')
+  [[ "$name" == "rust-example" || "$error_msg" =~ duplicate.*jobs_name_key ]] || exit 1;
```
bodymindarts, line 38: can't we add jitter to the name ( the "TEST_PREFIX" env var we discussed) instead of the cop-out via error.

bats/examples.bats
```diff
@@ -23,18 +32,18 @@ teardown_file() {
...
 
-  background cargo run --bin cala-ledger-example-rust > .rust-example-logs
+  background cargo run --bin cala-ledger-example-rust > .rust-example-logs 2>&1
 
-  sleep 5
+  job_count=$(cat .e2e-logs | grep 'Executing CalaOutboxImportJob importing' | wc -l)
```
bodymindarts, line 42: hmm this is no good - probably should rather have a delete or disable job so that there is always just 1 running... but okay for now I guess.

bats/examples.bats
```diff
@@ -23,18 +32,18 @@ teardown_file() {
...
+  retry 20 1 wait_for_new_import_job $job_count || true
+  sleep 1
 
   exec_graphql 'list-accounts'
-
-  name=$(graphql_output '.data.accounts.nodes[0].name')
-  [[ "$name" == "MY ACCOUNT" ]] || exit 1;
+  accounts_after=$(graphql_output '.data.accounts.nodes | length')
```
bodymindarts, line 47: Probably should change this assert to fetch account by name and assert by the jittered name instead of listing accounts.

bats/examples.bats
vindard: I was thinking this as well, but we didn't have a cancel job operation as yet and I wasn't sure what the behaviour would be if we had too many jobs running simultaneously.

We could also optimistically cancel/disable the job with this name and then run the create job to enable/re-enable it here. The `error_msg` check is more an interim solution.

bats/examples.bats
vindard: This is probably poorly named. Should be `job_executions_count`. I added it as a way to not arbitrarily, but to instead wait on a job execution before checking for the sync.

bats/examples.bats
vindard: Yup, interim solution as I wasn't sure if we should implement any new queries or mutations with the tests being the only forcing function. But maybe we're early enough that this is ok to drive the api design?

vindard:
Note: with the current HEAD at 97906dadd6595be8fbcc5886934f648b3317c126, if the server is restarted and a job already exists then this check (https://github.com/GaloyMoney/cala/blob/97906dadd6595be8fbcc5886934f648b3317c126/cala-server/src/job/registry.rs#L31) will fail since the registry isn't repopulated with that job type

bodymindarts:
> Note: with the current HEAD at [97906da](https://github.com/GaloyMoney/cala/commit/97906dadd6595be8fbcc5886934f648b3317c126), if the server is restarted and a job already exists then this check below will fail since the registry isn't repopulated with that job type
> 
> https://github.com/GaloyMoney/cala/blob/97906dadd6595be8fbcc5886934f648b3317c126/cala-server/src/job/registry.rs#L31

Ah hadn't considered that - thx

bats/examples.bats
bodymindarts: Yeah I think delete job and recreate would be good - should also fix the crash.

PR #43, opened 2024-05-11
chore: extension boilerplate

bodymindarts:


cala-server/schema.graphql
```diff
@@ -121,6 +121,7 @@ type JournalCreatePayload {
 type Mutation {
 	journalCreate(input: JournalCreateInput!): JournalCreatePayload!
 	importJobCreate(input: ImportJobCreateInput!): ImportJobCreatePayload!
+	test: Boolean!
```
bodymindarts, line 124: this will go away soon...

cala-server/src/extension/core.rs
```diff
@@ -0,0 +1,11 @@
+#[derive(async_graphql::SimpleObject, Default)]
+pub struct CoreMutationExtension {
+    hello: String,
```
bodymindarts, line 3: dummy

PR #45, opened 2024-05-12
feat: gql layer for tx_template

thevaibhav-dixit:


cala-ledger/src/tx_template/entity.rs
```diff
@@ -367,20 +367,20 @@ mod tests {
...
             .build()
             .unwrap()];
-        let new_journal = NewTxTemplate::builder()
+        let new_tx_template = NewTxTemplate::builder()
             .id(TxTemplateId::new())
             .code("CODE")
             .tx_input(
                 NewTxInput::builder()
-                    .effective("date('2022-11-01')")
+                    .effective("DATE('2022-11-01')")
```
bodymindarts, line 375: Why was this changed?

cala-server/src/graphql/convert.rs
```diff
@@ -82,6 +93,114 @@ impl From<cala_ledger::journal::JournalValues> for Journal {
...
+        Self {
+            name: value.name,
+            r#type: ParamDataType::from(value.r#type),
+            default,
+            description: value.description,
+        }
+    }
+}
+
+impl From<cala_ledger::tx_template::ParamDataType> for ParamDataType {
```
bodymindarts, line 174: I think async-graphql has an option to generate this conversion when you declare the GQL run type

cala-server/src/graphql/schema.rs
```diff
@@ -111,6 +112,105 @@ impl<E: MutationExtensionMarker> CoreMutation<E> {
...
+        let TxTemplateTxInput {
+            effective,
+            journal_id,
+            correlation_id,
+            external_id,
+            description,
+            metadata,
+        } = input.tx_input;
+        new_tx_input_builder
+            .effective(effective.as_ref().to_string())
```
bodymindarts, line 136: Why this as_ref().to_string() repeating pattern. Can probably improve this by accepting a trait

cala-server/src/graphql/tx_template.rs
```diff
@@ -0,0 +1,101 @@
+use async_graphql::*;
+
+use super::primitives::*;
+
+#[derive(SimpleObject)]
+pub(super) struct TxTemplate {
+    pub id: ID,
+    pub tx_template_id: UUID,
+    pub code: String,
+    pub params: Option<Vec<ParamDefinition>>,
```
bodymindarts, line 10: Why option? It’s a Vec so it can just be empty

thevaibhav-dixit: yep I was going to ask this question as to why we have this type in sqlx-ledger... I think we should change this to Vec<ParamDefinition>

cala-ledger/src/tx_template/entity.rs
thevaibhav-dixit: was testing something. committed by mistake

bodymindarts (approved):

PR #47, opened 2024-05-14
feat: account_create gql layer

thevaibhav-dixit:


cala-server/schema.graphql
```diff
@@ -26,6 +26,22 @@ type AccountConnection {
...
 
+input AccountCreateInput {
+	id: UUID
+	externalId: String
+	code: String!
+	name: String!
+	normalBalanceType: DebitOrCredit! = CREDIT
+	description: String
+	status: Status! = ACTIVE
+	tags: [TAG!]!
```
bodymindarts, line 37: tags should be optional on Input I think.

cala-server/src/graphql/account.rs
```diff
@@ -41,3 +41,23 @@ impl CursorType for AccountByNameCursor {
...
+    pub id: Option<UUID>,
+    pub external_id: Option<String>,
+    pub code: String,
+    pub name: String,
+    #[graphql(default)]
+    pub normal_balance_type: DebitOrCredit,
+    pub description: Option<String>,
+    #[graphql(default)]
+    pub status: Status,
+    pub tags: Vec<TAG>,
```
bodymindarts, line 56: probably put default here.

cala-server/src/graphql/primitives.rs
```diff
@@ -53,6 +72,18 @@ impl From<String> for TAG {
     }
 }
 
+impl AsRef<str> for TAG {
+    fn as_ref(&self) -> &str {
```
bodymindarts, line 76: why is this exposed?

cala-server/src/graphql/schema.rs
```diff
@@ -89,6 +89,46 @@ impl<E: MutationExtensionMarker> CoreMutation<E> {
...
+        builder
+            .id(id)
+            .name(input.name)
+            .code(input.code)
+            .normal_balance_type(input.normal_balance_type.into())
+            .status(input.status.into());
+        let tags = input
+            .tags
+            .into_iter()
+            .map(cala_ledger::Tag::try_from)
```
bodymindarts, line 114: much better!

cala-server/src/graphql/primitives.rs
thevaibhav-dixit: instead of making pub(super) i think it makes much more sense to call ```as_ref()``` during the try_from conversion

bodymindarts (approved):

PR #63, opened 2024-05-17
test: initial e2e test setup

thevaibhav-dixit:


bats/e2e.bats
```diff
@@ -0,0 +1,111 @@
...
+load "operations"
+load "helpers"
+
+setup_file() {
+  start_server
+}
+
+teardown_file() {
+  stop_server
+  stop_rust_example
```
bodymindarts, line 12: don't need this right?

bats/e2e.bats
```diff
@@ -0,0 +1,111 @@
...
+}
+
+@test "cala: journal create" {
+  variables=$(
+    jq -n \
+    '{
+        "input": {
+          "journalId": "822cb59f-ce51-4837-8391-2af3b7a5fc51",
```
bodymindarts, line 20: can we not hard code ids? That makes it harder to re-run the test

bats/e2e.bats
```diff
@@ -0,0 +1,111 @@
...
+          "name": "General Ledger",
+          "description": "Primary journal for Lava."
```
bodymindarts, line 22: lets not leak the name.

bats/gql/account.gql
```diff
@@ -0,0 +1,12 @@
+query account($accountId: UUID!, $journalId: UUID!, $currency: CurrencyCode!) {
```
bodymindarts, line 1: can we call this `account-with-balance`?

bats/operations.bash
```diff
@@ -0,0 +1,108 @@
+create_deposit_tx_template_variables() {
```
bodymindarts, line 1: I think this is not `operations` but `tx-templates`

Actually how about you just hard code this in the `tx-template-create.gql` file instead of putting all then inputs via bash / json.

You just need to inject the id I guess.

bodymindarts (approved):

PR #64, opened 2024-05-18
chore: balance gel layer

bodymindarts:
`Query { balance }` and `Account { balance }` fields to GQL layer.

Introduce DataLoader to re-use balance fetching code in both `fn`s.


cala-ledger/src/balance/repo.rs
```diff
@@ -12,14 +13,99 @@ use std::collections::{HashMap, HashSet};
...
+            JOIN cala_current_balances c
+            ON h.data_source_id = c.data_source_id
+            AND h.journal_id = c.journal_id
+            AND h.account_id = c.account_id
+            AND h.currency = c.currency
+            AND h.version = c.latest_version
+            JOIN cala_accounts a
+            ON c.data_source_id = a.data_source_id
+            AND c.account_id = a.id
+            WHERE c.data_source_id = '00000000-0000-0000-0000-000000000000'
```
thevaibhav-dixit, line 43: why this has been hardcoded ?

bodymindarts: Querying for synced entities is not on the critical path for our target use case. Hard coding this now to simplify. There are a few ways to deal with querying non-local entities and I'm not sure yet which approach is the right one.... will decide that later when relevant and restrict queries to local entities for now.

PR #106, opened 2024-06-12
feat: accountUpdate mutation

thevaibhav-dixit:

PR #110, opened 2024-06-13
feat: impl journalUpdate

thevaibhav-dixit:
REBASE once #106 gets merged!
NOTE: change the sync operation to sync_jounal_update


proto/ledger/outbox_service.proto
```diff
@@ -29,6 +29,7 @@ message CalaLedgerEvent {
     BalanceCreated balance_created = 12;
     BalanceUpdated balance_updated = 13;
     AccountUpdated account_updated = 14;
+    JournalUpdated journal_updated = 15;
```
bodymindarts, line 32: I think we should group these by resource:
```
account_created = 5;
account_updated = 6;
etc
```

bodymindarts (approved):

PR #116, opened 2024-06-17
feat: add sets query on account sets

thevaibhav-dixit:


cala-ledger/src/account_set/mod.rs
```diff
@@ -291,18 +291,52 @@ impl AccountSets {
     }
 
     #[instrument(
-        name = "cala_ledger.account_sets.find_where_account_is_member",
-        skip(self),
+        name = "cala_ledger.account_sets.find_where_member",
+        skip(self, member),
```
bodymindarts, line 295: don't think we should skip member... you can add `+ std::fmt::Debug` to the impl

bodymindarts (approved):

PR #117, opened 2024-06-17
docs: generate and update API reference for cala.sh

openoms:


bodymindarts:
I think we should only update docs when there is a release - not every commit to main. Is it possible to filter to only execute the action on tagged commits?

PR #146, opened 2024-06-28
chore: available balance

bodymindarts:


cala-ledger-core-types/src/balance.rs
```diff
@@ -13,16 +13,42 @@ pub struct BalanceSnapshot {
...
+            (other.modified_at, other.entry_id)
+        };
+
+        Self {
+            dr_balance: self.dr_balance + other.dr_balance,
+            cr_balance: self.cr_balance + other.cr_balance,
+            entry_id,
+            modified_at,
+        }
+    }
```
thevaibhav-dixit, line 53: maybe a unit test for this fn ?

PR #165, opened 2024-07-11
fix: members query in account set 

thevaibhav-dixit:
We were comparing `created_at` in `cala_account_set_member_account_sets` and `cala_account_set_member_accounts` with cursor created form  `created_at` in `cala_account_sets` and `cala_accounts`!


vindard:
Just did some testing with this branch. This is still pretty broken. It's grouping Accounts and then AccountSets in the returned list (sorted by cursor within each group).

- Account @ Time 6
- Account @ Time 4
- Account @ Time 2
- AccountSet @ Time 5
- AccountSet @ Time 3
- AccountSet @ Time 1

vindard:
Just did some testing with this branch. This is still pretty broken. It's grouping Accounts and then AccountSets in the returned list (sorted by cursor within each group).

For e.g, I'm getting (with time as proxy for cursor ordering):
- `Account @ Time 6`
- `Account @ Time 4`
- `Account @ Time 2`
- `AccountSet @ Time 5`
- `AccountSet @ Time 3`
- `AccountSet @ Time 1`

And if I paginate from Account @ Time 4 I'll get:
- `Account @ Time 2`
- `AccountSet @ Time 3`
- `AccountSet @ Time 1`

vindard:
I just added a branch to help with testing pagination. It seeds a list of accounts/accountSets under a `chart of accounts` account set that we can test pagination through

https://github.com/GaloyMoney/cala/tree/test-pagination

To initialize:
- `make reset-deps run-server`
- In a separate terminal `make run-tf`

To test, get the full list with this query, and then in a new tab choose different items to paginate from
```gql
accountSet(id: "00000000-0000-0000-0000-100000000000") {
    name
    members(first: 100) {
      edges {
        cursor
        node {
          ... on Account {
            __typename
            id
            name
          }
          ... on AccountSet {
            __typename
            id
            name
          }
        }
      }
    }
  }
}
```

## The bug

After drilling in further with this, the bug I found was that if you try to paginate from an AccountSet specifically (e.g. `(AccountSet #6) Deposits Control`), it will not show any neighbouring AccountSets as well. It will show all accounts after it, and all AccountSets after it that aren't its neighbours.

thevaibhav-dixit:
> Just did some testing with this branch. This is still pretty broken. It's grouping Accounts and then AccountSets in the returned list (sorted by cursor within each group).
> 
> For e.g, I'm getting (with time as proxy for cursor ordering):
> 
> * `Account @ Time 6`
> * `Account @ Time 4`
> * `Account @ Time 2`
> * `AccountSet @ Time 5`
> * `AccountSet @ Time 3`
> * `AccountSet @ Time 1`
> 
> And if I paginate **after** `Account @ Time 4` I'll get:
> 
> * `Account @ Time 2`
> * `AccountSet @ Time 3`
> * `AccountSet @ Time 1`
> 
> _We should probably figure out a way to test expected ordering as well_

is this not the expected behaviour ? we are getting all the accounts/ account sets that were created before Time @ 4

vindard:
> is this not the expected behaviour ? we are getting all the accounts/ account sets that were created before Time @ 4

There's no concept of time in the end-user pagination interface I think, we just use that as an implementation detail to do our ordering. If I was looking at the first list and paginating from Time 4 I would expect all items under it in the page, and ordered in the order the page was first presented:

- `Account @ Time 2`
- `AccountSet @ Time 5`
- `AccountSet @ Time 3`
- `AccountSet @ Time 1`

cala-ledger/src/account_set/repo.rs
```diff
@@ -116,21 +116,32 @@ impl AccountSetRepo {
                 member_created_at: last.created_at.expect("created_at not set"),
             });
         }
-        let ids = rows
+        let mut account_ids = Vec::new();
+        let mut account_set_ids = Vec::new();
+
+        for row in rows.into_iter() {
+            if let Some(member_account_id) = row.member_account_id {
+                account_ids.push((
```
bodymindarts, line 124: why not instantiate the proper type here?

thevaibhav-dixit: i was  not sure if we could sort the struct... But can try that

cala-ledger/tests/account_set.rs
```diff
@@ -261,3 +261,129 @@ async fn account_set_update() -> anyhow::Result<()> {
...
+        ret.entities[0].id.clone(),
+        AccountSetMemberId::from(set_two.id())
+    );
+    assert_eq!(
+        ret.entities[1].id.clone(),
+        AccountSetMemberId::from(account_one.id())
+    );
+
+    let after_cursor = AccountSetMemberCursor {
+        member_created_at: ret.entities[0].created_at,
```
bodymindarts, line 345: doesn't ret have the cursor? Why construct it>?

bodymindarts (approved):

cala-ledger/tests/account_set.rs
thevaibhav-dixit: yes it does have an end cursor. here i am creating cursor using the first element to better test

cala-ledger/src/account_set/cursor.rs
```diff
@@ -20,3 +20,9 @@ impl From<&AccountSetValues> for AccountSetByNameCursor {
 pub struct AccountSetMemberCursor {
     pub member_created_at: chrono::DateTime<chrono::Utc>,
 }
+
+impl From<chrono::DateTime<chrono::Utc>> for AccountSetMemberCursor {
```
bodymindarts, line 24: This should be from a domain type to another not from a primitive. so `From<AccountSetMember>`

bodymindarts (approved):

PR #194, opened 2024-08-16
chore: entries on account and account sets

thevaibhav-dixit:


cala-server/src/graphql/account.rs
```diff
@@ -139,6 +140,26 @@ impl Account {
         )
         .await
     }
+
+    async fn entries(
+        &self,
+        ctx: &Context<'_>,
+        from: Timestamp,
+        until: Option<Timestamp>,
```
thevaibhav-dixit, line 148: @vindard @bodymindarts this pr adds ranged query for `accounts` and `account_sets`. Should we also include a more generic paginated query as well ?

cala-server/src/graphql/entry.rs
```diff
@@ -0,0 +1,100 @@
...
+            direction: values.direction.into(),
+            description: values.description,
+            created_at: Timestamp::from(created_at),
+            modified_at: Timestamp::from(modified_at),
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+pub(super) struct EntryByCreatedAtCursor {
```
bodymindarts, line 60: Cursor needs to take `sequence` into account. `created_at` is not unique per account in case one account receives multiple entries in 1 transaction.

cala-ledger/src/entry/cursor.rs
```diff
@@ -0,0 +1,21 @@
+use chrono::{DateTime, Utc};
+use serde::{Deserialize, Serialize};
+
+use cala_types::primitives::EntryId;
+
+use super::Entry;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct EntryByCreatedAtCursor {
```
bodymindarts, line 9: need to add `sequence`

cala-server/src/graphql/entry.rs
```diff
@@ -0,0 +1,121 @@
...
+
+#[ComplexObject]
+impl Entry {
+    async fn transaction(&self, ctx: &Context<'_>) -> async_graphql::Result<Transaction> {
+        let transaction_id = TransactionId::from(self.transaction_id);
+        let ctx = ctx.data_unchecked::<DataLoader<LedgerDataLoader>>();
+        let transaction = ctx
+            .load_one(transaction_id)
+            .await?
+            .expect("a transaction should always exist for an entry");
```
bodymindarts, line 37: This might actually be `None` for "dummy" entries (like adding / removing accounts to account sets).

thevaibhav-dixit:
can bring it back when there is another trigger

PR #195, opened 2024-08-17
chore: add transactions query

thevaibhav-dixit:


cala-ledger/src/transaction/repo.rs
```diff
@@ -124,6 +125,50 @@ impl TransactionRepo {
...
+    ) -> Result<PaginatedQueryRet<Transaction, TransactionByCreatedAtCursor>, TransactionError>
+    {
+        let rows = sqlx::query_as!(
+            GenericEvent,
+            r#"
+        WITH transactions AS (
+          SELECT id, created_at
+          FROM cala_transactions
+          WHERE (created_at, id) < ($1, $2) OR ($1 IS NULL AND $2 IS NULL)
+          ORDER BY created_at DESC, id
```
bodymindarts, line 140: I would also order `id DESC` to be consistent... not a big deal though.

thevaibhav-dixit:
can bring it back when there is actual trigger for `transactionsBy__`

PR #305, opened 2025-01-22
feat: add AccountSet list_for_name query

vindard:


thevaibhav-dixit (approved):
LGTM... do we also want to expose this as a gql query?

PR #320, opened 2025-02-11
feat: add 'find_all_in_range' function

vindard:


cala-ledger/src/balance/repo.rs
```diff
@@ -188,6 +186,109 @@ impl BalanceRepo {
...
     }
 
+    pub(super) async fn find_range_all(
+        &self,
+        ids: &[BalanceId],
+        from: DateTime<Utc>,
+        until: Option<DateTime<Utc>>,
+    ) -> Result<HashMap<BalanceId, (Option<AccountBalance>, Option<AccountBalance>)>, BalanceError>
+    {
+        let mut query_builder = QueryBuilder::new(
```
bodymindarts, line 196: please use the - 'pass query args as arrays and use UNNEST in query' style. That way this query can use the `query!` improving type safety.

See for example the `find_for_update` function in the same file https://github.com/GaloyMoney/cala/blob/d680f1054fb0ad7992fa8f214f2d08556ea4c3e2/cala-ledger/src/balance/repo.rs#L204.

PR #347, opened 2025-03-17
feat: implement db operation rollback

vindard:
## Description

Needed to close a dp operation for a multi stage operation across a for loop when one iteration fails


cala-ledger/src/ledger_operation.rs
```diff
@@ -55,4 +55,10 @@ impl<'t> LedgerOperation<'t> {
         }
         Ok(())
     }
+
+    pub async fn rollback(self) -> Result<(), sqlx::Error> {
```
vindard, line 59: Do we need to consider outbox for anything here?

PR #366, opened 2025-03-26
feat: Add metadata to entries

### LINKED ISSUE #365: Add metadata to entry

jirijakes:
Previously only transaction templates and transactions had dedicated metadata but entries did not. This change adds equivalent metadata to entries as well.

Additionally, simple integration test is added that verifies transaction metadata were inserted. Ideally similar test would be added for entry metadata but I am unsure of how to do it at the moment.

Fixes #365.


thevaibhav-dixit (commented):
you probably need to run make sdl to bring in the schema changes

jirijakes:
@thevaibhav-dixit thanks! I ran it, let's see now.

bodymindarts (approved):

PR #373, opened 2025-03-30
chore: consistent filename across codebases

nicolasburtey:
necessitate to have consitent filename to extract events into a doc


bodymindarts (changes requested):
File name is correct AFAICT

PR #376, opened 2025-04-01
feat: Introduce BTC and USD currency constants

jirijakes:
Downstream projects frequently use `"USD".parse::<cala_ledger::Currency>().unwrap()` to obtain typed currency. This change adds currency constants for BTC and USD to substitute such in-place parsing.

Draft because still verifying in downstream whether this is sufficient.


cala-ledger-core-types/src/primitives.rs
```diff
@@ -156,6 +156,9 @@ impl Currency {
             Currency::Crypto(c) => c.code,
         }
     }
+
```
bodymindarts, line 159: Can you put the const at the top of the impl block

jirijakes: Done.

bodymindarts (approved):

PR #384, opened 2025-04-10
feat: add 'extend' to EntityEvents

vindard:


lib/es-entity/src/events.rs
```diff
@@ -73,6 +73,10 @@ where
         self.new_events.push(event);
     }
 
+    pub fn extend(&mut self, events: impl IntoIterator<Item = T>) {
+        self.new_events.extend(events);
+    }
+
```
thevaibhav-dixit, line 79: what is the usecase for this ?

thevaibhav-dixit (approved):

lib/es-entity/src/events.rs
vindard: Adding multiple events in one call

bodymindarts (approved):

PR #385, opened 2025-04-11
feat: Add ability to query transactions by template

jirijakes:
This would allow to somehow collect templates of interest (e. g. by selecting specific template codes or by a glob-like expression like `*DEPOSIT*`) and call `Transactions::list_for_template_ids()` with their IDs.


cala-ledger/src/transaction/mod.rs
```diff
@@ -10,10 +10,12 @@ use std::collections::HashMap;
 
 #[cfg(feature = "import")]
 use crate::primitives::DataSourceId;
+use crate::primitives::TxTemplateId;
```
thevaibhav-dixit, line 13: why not include this with rest of the imports in L 14

jirijakes: Ah, yeah, it could have been there. I only noticed the feature-gated `crate::primitives` above, did not see the one on L14.

PR #404, opened 2025-05-05
feat: effective balances

bodymindarts:


jirijakes:
This is nice! Reviewed code, played with tests, checked content in DB and did some pencil tracing to verify the implementation. I have one suggestion for improving readability (see review comment above).

jirijakes (approved):

PR #405, opened 2025-05-06
feat: add 'has_activity' function to BalanceSnapshot

vindard:


thevaibhav-dixit (commented):
I am not sure if we need this in cala. probably @bodymindarts knows better

bodymindarts:
Closing as not relevant for core Cala.

PR #437, opened 2025-06-04
fix: sequence calculation in persist_events_batch

thevaibhav-dixit:


lib/es-entity-macros/src/repo/persist_events_batch_fn.rs
```diff
@@ -134,7 +134,7 @@ mod tests {
                     all_serialized.extend(serialized);
                     all_types.extend(types);
                     all_ids.extend(std::iter::repeat(id).take(n_events));
-                    all_offsets.extend((offset..).take(n_events).map(|i| i as i32));
+                    all_sequences.extend((offset..).skip(1).take(n_events).map(|i| i as i32));
```
bodymindarts, line 137: how about ` let offset = events.len_persisted() + 1;` instead of `skip(1)`?

PR #471, opened 2025-07-04
feat: void transaction 

thevaibhav-dixit:


cala-ledger/src/entry/mod.rs
```diff
@@ -96,6 +96,47 @@ impl Entries {
...
+        new_tx_id: TransactionId,
+    ) -> Result<Vec<NewEntry>, EntryError> {
+        let entries = self.list_for_transaction_id(original_tx_id).await?;
+
+        let new_entries = entries
+            .into_iter()
+            .map(|entry| {
+                let value = entry.into_values();
+                let mut units = value.units;
+                units.set_sign_negative(true);
```
bodymindarts, line 111: I think this is wrong - we need to flip the sign not set it negative. (It might already be negative).

cala-ledger/src/ledger/mod.rs
```diff
@@ -222,6 +222,82 @@ impl CalaLedger {
...
+                new_tx_id,
+                new_entries.iter().map(|entry| entry.id),
+            )
+            .await?;
+
+        let span = tracing::Span::current();
+        span.record("transaction_id", transaction.id().to_string());
+        span.record("external_id", &transaction.values().external_id);
+
+        let entries = self.entries.create_all_in_op(db, new_entries).await?;
```
bodymindarts, line 267: these is something inconsistent here:
```
  let new_entries = self
            .entries()
```
there you are using `entries()` and here `entries`. Also why do we need to go to `entries` twice? Can't we create + persist in 1 round trip to that service?

cala-ledger/src/transaction/mod.rs
```diff
@@ -45,6 +45,22 @@ impl Transactions {
...
+        original_tx_id: TransactionId,
+        new_tx_id: TransactionId,
+        entry_ids: impl IntoIterator<Item = EntryId>,
+    ) -> Result<Transaction, TransactionError> {
+        let mut original_tx = self.repo.find_by_id(original_tx_id).await?;
+        let new_tx = original_tx.void(new_tx_id, entry_ids.into_iter().collect(), db.op().now())?;
+
+        self.repo.update_in_op(db.op(), &mut original_tx).await?;
+        let transaction = self.repo.create_in_op(db.op(), new_tx).await?;
+        db.accumulate(transaction.last_persisted(1).map(|p| &p.event));
```
bodymindarts, line 60: also need to accumulate the `update_in_op` / `original_tx` event.

PR #502, opened 2025-08-25
feat: implement account-set-aware velocities

vindard:


cala-ledger/src/velocity/balance/mod.rs
```diff
@@ -72,14 +78,26 @@ impl VelocityBalances {
...
+                    .iter()
+                    .map(AccountId::from)
+                    .collect::<Vec<_>>(),
+            );
+            for account_id in ids {
+                let Some((_, controls)) = controls.get(&account_id) else {
+                    continue;
+                };
+                for control in controls.iter() {
+                    if !control.needs_enforcement(&ctx)? {
```
vindard, line 98: This breaks since context is unable to fetch account values for account set id

bodymindarts: there are only account values cached in the context if there is actually a control.
Also `let ctx = context.context_for_entry(entry);` will have to receive the `account_id` as well since we may be enforcing on an account that is not the `entry.account_id`.

bodymindarts: I have fixed this instance [here](https://github.com/GaloyMoney/cala/pull/502/commits/08d1f672df32a8f076daf22f4ae07da986d036a4) but the same thing needs to be applied at the snapshot creation loop.

cala-ledger/src/velocity/mod.rs
```diff
@@ -157,18 +157,17 @@ impl Velocities {
         account_ids: &[AccountId],
         account_set_mappings: &HashMap<AccountId, Vec<AccountSetId>>,
     ) -> Result<(), VelocityError> {
-        let mut unique_ids = account_ids.iter().copied().collect::<HashSet<_>>();
```
vindard: For this I was thinking there could be cases where 2 accounts in a transaction fall under the same account set in a chart of accounts for e.g., which would mean we'd get multiple instances of every account id in the chart tree and any statement trees that attach to the chart. Not sure if this is a concern for controlling the number of ids we eventually pass to the sql query?

cala-ledger/src/velocity/mod.rs
```diff
@@ -107,7 +106,7 @@ impl Velocities {
     pub async fn attach_control_to_account(
```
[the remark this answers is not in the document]
vindard: I wonder if it's worth it to do separate use-cases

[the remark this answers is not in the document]
jirijakes: That might be even better.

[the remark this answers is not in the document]
vindard: https://github.com/GaloyMoney/cala/pull/502/commits/65923da8aa0bb30250d600e6b9bada250befca33

PR #507, opened 2025-08-29
feat: implement `Date` and `Timestamp` comparision in CEL interpreter

vindard:
## Description

This PR extends CEL interpreter and uses new functionality in an effective-date-conditioned velocity control


cala-ledger/tests/velocity.rs
```diff
@@ -191,3 +228,147 @@ async fn create_control_on_account_set() -> anyhow::Result<()> {
...
+    let velocity = cala.velocities();
+
+    let debit_limit = account_closing_limit(velocity, "DEBIT").await?;
+    let credit_limit = account_closing_limit(velocity, "CREDIT").await?;
+
+    let control = NewVelocityControl::builder()
+        .id(VelocityControlId::new())
+        .name("Account Closing")
+        .description("test")
+        .condition("context.vars.transaction.effective <= date('2024-12-31')")
```
bodymindarts, line 257: wouldn't it make more sense to have something like - effective is _before the current month_ or something else dynamic like that. Wouldn't want to have to create a new velocity every month.

jireva: Yes, I think the expectation here would be that closing the books would be an explicit action.  I've sent a message to Guillermo to ask if some automatic close after the grace period would be ok to start with.

PR #565, opened 2025-09-30
feat(nodejs): add bindings for tx_templates

sutt0n:
- added `tx_template` bindings
  - `find_by_code` method
  - `create` method (with `transaction` property validation)
- bats tests
  - separated nodejs and rust examples
  - nodejs example checks for accounts and created `tx_template` by its code
- tsx added to `flake.nix` to ensure reproducibility

### DIFF
--- a/bats/nodejs-examples.bats
+++ b/bats/nodejs-examples.bats
@@ -1,0 +1,70 @@
+#!/usr/bin/env bats
+
+load "helpers"
+
+setup_file() {
+  reset_pg_and_restart_server
+}
+
+teardown_file() {
+  stop_server
+  stop_nodejs_example
+}
+
+reset_pg_and_restart_server() {
+  stop_server
+  reset_pg
+  PG_CON=$PG_CON_EXAMPLE start_server
+}
+
+@test "nodejs: entities sync to server" {
+  reset_pg_and_restart_server
+
+  exec_graphql 'list-accounts'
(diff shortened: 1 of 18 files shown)


bats/nodejs-examples.bats
```diff
@@ -0,0 +1,70 @@
...
+
+teardown_file() {
+  stop_server
+  stop_nodejs_example
+}
+
+reset_pg_and_restart_server() {
+  stop_server
+  reset_pg
+  PG_CON=$PG_CON_EXAMPLE start_server
```
bodymindarts, line 17: why is the PG_CON here different than in setup_file?

bodymindarts: That breaks the point of this test.

The scenario is that the example uses its own DB and the server syncs via the outbox.
This is testing implicitly testing the federated feature.
The nodejs example should work the same as the rust example.

The server and the nodejs code using the same DB defeats the purpose of the test (obviously the data will be identical).

sutt0n: I thought that was odd... Alright, after carefully re-reading the test and playing around with the outbox server and handler a bit, I've pushed an update -- this test was absolutely written on my misunderstanding of the functionality. I had completely overlooked the outbox sync entirely...

I've reverted some changes back, and _actually_ fired off the outbox handler in the NodeJS example to sync the entities and test the federated feature.

These changes are pushed.

PR #704, opened 2026-04-10
feat(balance): publish effective balance events to outbox

Lakshyyaa:
## Summary
- Add `EffectiveBalanceSnapshot` type and `EffectiveBalanceCreated`/`EffectiveBalanceUpdated` outbox event variants
- Thread `OutboxPublisher` through to `EffectiveBalanceRepo`
- Publish events after inserting into `cala_cumulative_effective_balances`, matching the existing `BalanceRepo` pattern
- Refactor `insert_new_snapshots` to accept `Vec<EffectiveBalanceSnapshot>` with a `new_effective_snapshots()` conversion method, consistent with `BalanceRepo::new_snapshots()`

## Why
`cala_cumulative_effective_balances` was the only CALA table (out of 6 used by downstream reports) that didn't publish to the outbox. This closes that gap so a downstream relay can forward these events.

## Test plan
- [x] `cargo check` passes
- [ ] Existing tests pass
- [ ] Verify outbox events are published after effective balance inserts

### DIFF
--- a/cala-ledger/src/balance/effective/mod.rs
+++ b/cala-ledger/src/balance/effective/mod.rs
@@ -7,12 +7,14 @@ use std::collections::{HashMap, HashSet};
 use tracing::instrument;
 
 use cala_types::{
-    balance::{BalanceAmount, BalanceSnapshot},
+    balance::{BalanceAmount, BalanceSnapshot, EffectiveBalanceSnapshot},
     entry::EntryValues,
     primitives::*,
 };
 
-use crate::primitives::JournalId;
+use crate::{outbox::OutboxPublisher, primitives::JournalId};
+
+use data::EffectiveBalanceData;
 
 use super::{account_balance::*, error::BalanceError, snapshot::UNASSIGNED_ENTRY_ID};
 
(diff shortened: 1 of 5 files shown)


cala-ledger/src/balance/effective/mod.rs
```diff
@@ -343,10 +345,38 @@ impl EffectiveBalances {
...
+        let new_balances = Self::new_effective_snapshots(journal_id, all_data);
         self.repo
-            .insert_new_snapshots(op, journal_id, all_data)
+            .insert_new_snapshots(op, journal_id, new_balances)
             .await?;
 
         Ok(())
     }
+
+    fn new_effective_snapshots(
```
sandipndev, line 356: I don't think this refactoring Is related to this PR

sandipndev: Can add the publish all right at the end of the for loop here. Data structure change here seems unnecessary. https://github.com/GaloyMoney/cala/pull/704/changes#diff-49490a4e5aa32e9d913a4c82f40224f11827d5e88c08d77429a4fce9fda5dd32L756

Lakshyyaa: added this to keep it consistent with the `BalanceSnapshots`, : https://github.com/GaloyMoney/cala/blob/main/cala-ledger/src/balance/mod.rs#L161-L172
can remove if u think it is not necessary

sandipndev (approved):

PR #718, opened 2026-05-01
refactor!: replace custom CEL evaluator with cel-rust

nicolasburtey:
## Summary
- replace the custom CEL evaluator path in cala-cel-interpreter with cel-rust's Program/Context execution
- add compatibility conversions between Cala CelValue and cel-rust Value, including opaque Decimal and Uuid wrappers
- re-register Cala-specific CEL functions: date, uuid, decimal, decimal.Add, and timestamp format
- remove the retired cala-cel-parser crate, generated LALRPOP parser, stale interpreter package modules, and release-script references to the parser crate
- adjust date default validation for the timestamp-backed date representation

## Refresh (2026-07-29)
- merged latest `main` (was 73 commits behind / conflicting): dropped `cala-cel-parser` (this PR) and `cala-tracing` (removed on main in #727), kept main's newer deps (es-entity 0.11.9, job 0.6.35, obix 0.4.2)
- **bumped `cel` 0.13.0 → 0.14.1** (latest)
- **added a global compiled-program cache** (`#[cached(max_size = 10000)]` on `Program::compile`): `CelExpression`s are re-created from source on every DB read (velocity controls are deserialized per transaction), and cel-rust's ANTLR compile is ms-scale expensive vs the old memoized lalrpop parser. Without this cache, velocity paths regressed.
- added `cala-cel-interpreter/benches/cel_eval.rs` criterion micro-benchmarks

## Verification
- `cargo check --workspace --all-targets` ✅
- `cargo test -p cala-cel-interpreter` ✅ (8/8)
- CI: Check Code ✅, Integration Test ✅, security_audit ✅, performance-report ✅

## Performance
Compared `main` (`4b3df497`, custom evaluator) against this branch (`f22c9bb5`, cel-rust 0.14.1 + compile cache). Criterion, 100 samples, fresh Postgres per run.

### Pure CEL micro-benchmark (`cel_eval.rs`, same bench code on both branches)
| Workload | `main` (custom) | cel-rust 0.14.1 | Delta |
|---|---:|---:|---:|
| parse cache hit | 130.6 ns | 50.9 ns | **−61%** |
| eval_logic (`a ? b : c`) | 87.5 ns | 74.2 ns | **−15%** |
| eval_lookup (`params.hello.world`) | 329.7 ns | 362.8 ns | +10% |
| eval_decimal_eq (velocity-style) | 592 ns | 792 ns | +34% |
| eval_date_fn | 139.5 ns | 224 ns | +60% |

cel-rust is faster on parse-cache hits and pure logic, but 10–60% slower per function call (`decimal()`, `date()`). At nanosecond scale this is negligible next to a 5–20 ms DB transaction.

### DB-backed `cala-perf` ledger benchmarks (vs `main` baseline)
| Benchmark | Delta | Significant? |
|---|---:|---|
| post_multi_layer_transaction | **−17%** | ✅ p<0.01 |
| post_simple_transaction_with_effective_balances | **−63%** | ✅ p<0.01 (re-parses CEL per txn; uncached ANTLR compile cost ~7 ms/txn before the cache fix — reproduced across 2 runs, though the `main` baseline was the first/cold run, so treat magnitude with caution) |
| post_simple_transaction_with_five_account_sets | **−6%** | ✅ p=0.03 |
| post_simple_transaction | ~0% | no |
| post_simple_transaction_with_velocity | +4% | ⚠️ p<0.01 |
| post_simple_transaction_with_hit_velocity | +2% | no (p=0.12) |
| post_and_batch_recalculate_ec_account_set | +7% | ⚠️ p<0.01 |
| others (one/ec account set, skipped velocity, recalculate ec) | ~0% | no |

**Net:** clear win on CEL-heavy transaction paths (multi-layer, account sets, effective balances); small residual cost (~2–7%) on velocity-enforcement paths where cel-rust per-call overhead is slightly higher than the old tree-walking evaluator. Possible follow-up: avoid re-parsing `decimal('...')` string literals on every evaluation.

## Not run
- OTel/Jaeger runtime tracing.

---

> [!NOTE]
> **High Risk**
> High risk because it replaces the core CEL parsing/execution engine and rewires type/function handling, which can change evaluation semantics and error behavior across templates/params.
>
> **Overview**
> Switches `cala-cel-interpreter` from the in-repo `cala-cel-parser` + custom AST evaluator to `cel::Program`/`cel::Context` execution, compiling expressions once and executing them against a registered runtime context.
>
> Reimplements Cala builtins on top of `cel` (e.g. `date`, `uuid`, `decimal`, `decimal.Add`, timestamp `format`) and adds bidirectional conversions between Cala `CelValue`/`CelKey` and `cel::objects::Value` using opaque wrappers for Decimal/Uuid.


cala-cel-interpreter/src/builtins/mod.rs
```diff
@@ -1,40 +1,118 @@
...
 }
 
-#[instrument(name = "cel.builtin.uuid", skip_all, level = "debug", err(level = tracing::Level::WARN))]
-pub(crate) fn uuid(args: Vec<CelValue>) -> Result<CelValue, CelError> {
-    let s: Arc<String> = assert_arg(args.first())?;
-    Ok(CelValue::Uuid(
-        s.parse()
-            .map_err(|e| CelError::UuidError(format!("{e:?}")))?,
-    ))
+pub(crate) fn decimal_add(Arguments(args): Arguments) -> Result<Value> {
```
bodymindarts, line 77: **[blocker] Decimal loses ordering and arithmetic operators — silent runtime regression.**

The old evaluator supported `+`, `-`, `*` and `<`, `<=`, `>`, `>=`, `==`, `!=` on `Decimal×Decimal` (old `interpreter.rs` `evaluate_arithmetic`/`evaluate_relation`). With `Decimal` as `Value::Opaque`, cel-rust 0.14.1 supports **only `==`/`!=`** — I verified in the crate source: `OpaqueVal`'s `Val` impl provides `equals` only; `PartialOrd for Value` returns `None` for opaques, so `<` → `NoSuchOverload` and `+` → `UnsupportedBinaryOperator` **at evaluation time**. Only `decimal.Add` is re-registered; there is no `Sub`/`Mul` or any comparison escape hatch.

Because these fail at eval time (parse + deserialization succeed), a stored velocity condition like `context.vars.entry.units > decimal('100')` — which worked on main — starts rejecting every transaction that triggers it. Current in-repo tests and lana-bank only use `decimal('...')` literals with `==`, so nothing breaks *today*, but cala-ledger is a published crate and this is exactly the kind of expression a velocity condition invites.

Ask (pick one, but explicitly):
1. Add `decimal.Sub` / `decimal.Mul` / `decimal.Cmp` (or `Gt`/`Lt`) package functions and document that operators don't apply to decimals, or
2. Pursue comparable-opaque support upstream in cel-rust, or
3. Document the restriction as a breaking change in the CHANGELOG with migration guidance.

Either way this needs a release-note entry — silent is the only wrong option here.

cala-cel-interpreter/src/value.rs
```diff
@@ -92,6 +100,98 @@ impl From<CelMap> for CelValue {
...
+                CelValue::List(Arc::new(res))
+            }
+            Value::Int(i) => CelValue::Int(i),
+            Value::UInt(u) => CelValue::UInt(u),
+            Value::Float(f) => CelValue::Double(f),
+            Value::String(s) => CelValue::String(s),
+            Value::Bytes(b) => CelValue::Bytes(b),
+            Value::Bool(b) => CelValue::Bool(b),
+            Value::Duration(d) => CelValue::String(Arc::new(format!("{d:?}"))),
+            Value::Timestamp(ts) => CelValue::Timestamp(ts.with_timezone(&Utc)),
```
bodymindarts, line 166: **[blocker] Date → Timestamp changes persisted JSON formats — including velocity window partition keys (DB match keys).**

Since `date()` now returns `Value::Timestamp` and this branch maps every timestamp back to `CelValue::Timestamp`, `CelValue::Date` is unreachable on the result path. Two persisted-data consequences:

1. **Velocity windows** (the sharper one): `AccountVelocityLimit::window_for_enforcement` serializes each partition-key CEL result to JSON, and that `Window` is part of `VelocityBalanceKey` — matched against the persisted `partition_window` in `find_for_update`. A date-valued partition key (e.g. `context.vars.transaction.effective` for a daily limit) changes from `"2022-10-10"` to `"2022-10-10T00:00:00+00:00"`, so post-upgrade windows **never match pre-upgrade rows** → tracked balances silently restart from zero → limits under-enforce. lana-bank currently uses `window(vec![])` everywhere so it's unaffected, but for any cala user with date-partitioned windows this is silent data-compat breakage with no error anywhere.
2. **Transaction/entry metadata** (Bugbot's point, confirmed): `metadata.try_evaluate::<serde_json::Value>()` now stores RFC3339 strings where `%Y-%m-%d` strings were stored before — downstream report/consumer drift.

I don't see a clean in-crate fix — an opaque `CelDate` wrapper would preserve serialization but break `date(...) >= effective` comparisons (which lana-bank's closing lock needs, and which work today precisely *because* dates became timestamps). So the ask is: make this an explicit, documented decision — CHANGELOG entry, release note, and ideally a migration note for date-partitioned windows (e.g. "velocity balances keyed on date partitions will re-window"). If any known deployment uses date-valued partition keys, that needs a data migration, not a footnote.

cala-cel-interpreter/src/value.rs
```diff
@@ -92,6 +100,98 @@ impl From<CelMap> for CelValue {
...
+                }
+                CelValue::List(Arc::new(res))
+            }
+            Value::Int(i) => CelValue::Int(i),
+            Value::UInt(u) => CelValue::UInt(u),
+            Value::Float(f) => CelValue::Double(f),
+            Value::String(s) => CelValue::String(s),
+            Value::Bytes(b) => CelValue::Bytes(b),
+            Value::Bool(b) => CelValue::Bool(b),
+            Value::Duration(d) => CelValue::String(Arc::new(format!("{d:?}"))),
```
bodymindarts, line 165: **[nit]** `Value::Duration(d)` → `CelValue::String(format!("{d:?}"))` produces chrono's Debug format (`TimeDelta { secs: 3600, nanos: 0 }`-style), which isn't useful to any consumer and will end up in metadata JSON if someone evaluates `duration('1h')` (now a *gained* std function from cel-rust). Either add a proper `CelValue::Duration` variant or return `CelError::Unexpected` — a stringly Debug dump is the worst of both.

cala-cel-interpreter/src/error.rs
```diff
@@ -50,3 +50,22 @@ pub enum CelError {
...
     EvaluationError(String, Box<Self>),
 }
+
+impl From<cel::ExecutionError> for CelError {
+    fn from(error: cel::ExecutionError) -> Self {
+        match error {
+            cel::ExecutionError::UndeclaredReference(name) => {
+                CelError::UnknownIdent(name.to_string())
+            }
+            cel::ExecutionError::NoSuchKey(name) => CelError::UnknownIdent(name.to_string()),
```
bodymindarts, line 60: **[question]** This mapping is where a real behavior change surfaces: in the old evaluator, a missing key on a *top-level* context map returned `Null` (old `evaluate_member` called `map.get(name)` without a `contains_key` guard for `ContextItem::Value` maps), while nested missing keys errored. cel-rust returns `NoSuchKey` at **every** depth. So e.g. a template metadata expression referencing `params.optionalParam` (defined param, not provided, no default → key absent from the params map) used to evaluate to `Null` and now fails the transaction post. More spec-compliant, but: are we confident no stored templates rely on the old `Null` propagation? This is another eval-time-only change that a stored-expression scan would catch.

**[nit]** Several `CelError` variants are now dead code (`UnknownPackage`, `UnknownAttribute`, `IllegalTarget`, `WrongArgumentType`, `ChronoParseError`, `UuidError`, `DecimalError`, `TimestampError`) — worth pruning, or keeping deliberately for API compat with a comment.

cala-ledger-core-types/src/param.rs
```diff
@@ -32,6 +32,13 @@ impl ParamDataType {
             String if *self == ParamDataType::String => Ok(value),
             Map if *self == ParamDataType::Json => Ok(value),
             Date if *self == ParamDataType::Date => Ok(value),
+            Timestamp if *self == ParamDataType::Date => {
+                if let CelValue::Timestamp(ts) = value {
+                    Ok(CelValue::Date(ts.date_naive()))
+                } else {
```
bodymindarts, line 38: **[concern]** This coercion covers *provided* params, but default expressions bypass it: `Params::into_context` inserts `expr.evaluate(&ctx)?` **uncoerced**, so a `Date`-typed param defaulted with `date()` now lands in the CEL context as a `Timestamp`, not a `Date`. The typed extraction paths are fine (`TryFrom<CelResult> for NaiveDate` accepts `Timestamp`), but if that param feeds `metadata` or a window partition key, you get the RFC3339-vs-`%Y-%m-%d` drift flagged in `value.rs` — via a path (`date()` defaults, used by `cala-perf` and the test helpers) that's very common in real templates. Consider running defaults through `coerce_value` too, so a `Date` param is a `CelValue::Date` in the context regardless of where its value came from.

**[nit]** Pre-existing, but while here: `ParamDataType::Timestamp` still has no accepting arm at all, so a `Timestamp`-typed param always fails coercion.

cala-cel-interpreter/src/interpreter.rs
```diff
@@ -516,62 +184,14 @@ mod tests {
 
     #[test]
     fn has_macro_with_map() {
```
bodymindarts, line 228: **[suggestion]** Two test asks for an engine swap of this size:

1. The old suite's nested `has()` cases (`has(config.database.settings.maxConnections)` existing → `true`, missing leaf → `false`) were dropped here — cel-rust handles both (test-select on maps), so restoring them is free coverage of the macro path lana-bank's closing conditions depend on.
2. There's no differential corpus locking in behavior across the swap. A table-driven test of representative expressions → expected `CelValue` **and expected `serde_json::Value` serialization** (velocity-style decimal `==`, `date(...) >= timestamp` comparisons, `has()` guard chains, param-default `date()`, window-style partition values) would have surfaced the Date-serialization and decimal-operator changes before review, and will guard the next `cel` version bump — 0.13 → 0.14 already happened mid-PR; more bumps will follow.

bodymindarts (changes requested):
Reviewed the full engine swap (old lalrpop parser + tree-walker vs `cel` 0.14.1), including the cel-rust 0.14.1 source to verify operator semantics for `Value::Opaque`, and the downstream CEL surface (velocity conditions/windows, tx-template params/metadata, lana-bank's month-closing lock conditions).

## Overall
This is a good direction and the execution is careful: the compile cache restores parity with the old memoized parser, the benchmark work (micro + DB-backed, both branches) is exactly what a change like this needs, and cel-rust fixes real defects in the old evaluator — `1/2`, `1%2`, and `a[0]` previously hit `unimplemented!()` and **panicked the process**; `==` on strings/bools/uuids didn't work at all; `||`/`&&` now have CEL-spec error absorption. Integration tests (velocity `has()` guards, `date(...) >= effective` closing conditions) passing is a strong signal, and I verified lana-bank's monthly-closing CEL conditions evaluate correctly under the new semantics.

That said, a full evaluator replacement in a **published financial ledger library** ships two silent semantic changes that I think need explicit resolution before merge — both fail at *evaluation time*, not parse time, so stored expressions in downstream DBs deserialize fine and then behave differently at runtime. Details in the inline comments; requesting changes on these three:

### Blocking asks
1. **Decimal loses ordering + arithmetic operators** (inline at `builtins/mod.rs`). Old evaluator supported `+ - *` and `< <= > >=` on `Decimal×Decimal`; `Value::Opaque` in cel-rust 0.14.1 supports only `==`/`!=` (I checked `OpaqueVal`'s `Val` impl — it implements `equals` only, no comparer/adder). Needs a decision: helper functions, upstream support, or an explicit documented restriction.
2. **`date()` results are now timestamps in JSON output** (inline at `value.rs`) — this changes persisted formats, including **velocity window partition keys**, which are DB match keys. Needs an explicit compat decision + release note.
3. **Branch is `CONFLICTING` with main** (main is at 0.19.x / es-entity 0.11.11 with several releases since the last merge) — needs another merge, and the green CI predates it.

### Non-blocking
- Missing-key semantics changed for top-level context maps: `params.missing` returned `Null` in the old evaluator (only *nested* missing keys errored); cel-rust errors with `NoSuchKey` at every depth. More spec-compliant, but worth confirming no stored templates rely on `Null` propagation (question inline at `error.rs`).
- Test coverage: the old nested `has()` test cases were dropped, and there's no differential corpus locking in JSON serialization shapes (suggestion inline).
- Given both blocking items are behavior changes rather than bugs, a pre-deploy scan of stored expressions (`SELECT` all CEL sources from tx-template/velocity tables, compile + eval against representative contexts) would de-risk rollout for existing deployments considerably.

Nice work on the perf methodology — the honest annotation of the cold-run caveat on the −63% number is appreciated.

bodymindarts:
Follow-up on the **Date → Timestamp serialization** finding (review blocker #2), with deployment evidence: I scanned all four deployed lana-bank runtime DBs (galoy-staging `main` + `canary`, volcano-qa `main` + `vanilla`) for stored CEL surfaces:

- **Velocity windows**: 0 of 20 stored velocity limits across the four DBs have a non-empty `window`; every persisted `partition_window` in `cala_velocity_current_balances` is `{}` → **no window-key migration risk for any known deployment**.
- **Conditions**: the only stored CEL condition anywhere is the monthly-closing lock (`has()` guard chain + `date(...) >= context.vars.transaction.effective`), which I verified evaluates correctly under cel-rust (timestamp ordering + spec-compliant `||` error absorption).
- **Templates**: `effective` is always `params.effective` (Date-typed, always provided → String→Date coercion path, unchanged), metadata is always `params.meta` (Json-typed, content produced Rust-side — never `date()` output), **zero param defaults** stored, and all `units` expressions are bare `params.X` — no decimal arithmetic or ordering anywhere.
- `meta` is provided at every lana-bank posting call site, so the Null→`NoSuchKey` change for absent params is also not hit.

**Net effect on my review**: for known deployments, both semantic-drift blockers are *latent*, not active — no data migration needed. The remaining asks reduce to: (1) CHANGELOG/release-note entries for external crates.io users (decimal operator loss, date JSON format, missing-key errors, mixed date/timestamp comparisons now silently compare at ns precision instead of erroring — recommend documenting `date(x)` as the day-granularity normalizer), (2) the `decimal.Sub/Mul/Cmp` helper decision, and (3) the rebase (branch is still conflicting with main).

bodymindarts:
Pushed two commits to get this merge-ready (per Justin's request):

1. **`316d3b50` — Merge `main`** — resolved conflicts: dropped `cala-cel-parser` (this PR), kept main's `0.19.3-dev` versions + newer deps (es-entity 0.11.11, job 0.6.36, obix 0.4.3), reconciled `Cargo.lock` with `cargo update -w`. Branch is now mergeable.
2. **`23e3d838` — `feat(cel): add decimal.Sub, decimal.Mul and decimal.Cmp`** — resolves review blocker #1 (decimal operator loss). `decimal.Cmp` returns `-1/0/1` so it composes with native int comparisons (`decimal.Cmp(a, b) > 0` ⇔ `a > b`). All three use **checked arithmetic** (overflow → function error instead of a `rust_decimal` panic; `decimal.Add` switched to `checked_add` too) and coerce string/int args like `decimal()` does. Tests added (10/10 passing locally via `nix develop -c cargo test -p cala-cel-interpreter`).

Remaining from my review (non-blocking after the [deployed-DB scan](https://github.com/GaloyMoney/cala/pull/718#issuecomment-5141378395)): a release-note/CHANGELOG mention of the behavior changes for external crates.io users — date JSON format (RFC3339 instead of `%Y-%m-%d`), missing-key errors instead of `Null`, native decimal operators → `decimal.*` functions, and mixed date/timestamp comparisons no longer erroring. Will approve once CI is green.

bodymindarts (approved):

bodymindarts:
Approving - awesome that we can do this now.

PR #744, opened 2026-07-13
feat(ledger)!: require effective date when voiding a transaction

nicolasburtey:
## Motivation

`void_transaction` hardcoded the reversal's effective date to the void day (`Transaction::void` used `created_at.date_naive()`), with no way to override it. Voiding a backdated transaction therefore left the effective-balance timeline asymmetric: the original delta stayed on its effective date while the reversal landed on the void date — both periods' activity was permanently distorted. Standard correction accounting wants same-period reversal while the period is open, current-period reversal once it is closed.

Rather than keeping the void-day default alongside a new variant, the effective date is now a required parameter: dating a reversal is an accounting decision the caller should make explicitly, and a silent "today" default is exactly the footgun this removes.

## Changes

- `CalaLedger::void_transaction` / `void_transaction_in_op` take an explicit `effective: NaiveDate` for the reversal entries.
- `Transaction::void` rejects an effective date earlier than the original transaction's with a new `TransactionError::VoidEffectiveBeforeOriginal` — a reversal dated before the thing it reverses is never coherent.
- `Transactions::create_voided_tx_in_op` gains the same required parameter.

**BREAKING**: all three methods change signature. Callers wanting the previous behaviour pass their clock's current date.

Everything downstream is untouched on purpose: the reversal flows through the normal posting path, so a backdated void gets the same effective-balance cascade (rollforward of later cumulative snapshots) as any backdated post, and velocity controls evaluate `transaction.effective` from the actual transaction values — meaning period-close controls (e.g. lana-bank's closing velocity control) correctly block voids into closed periods.

## Tests

- Unit: void on the original effective date; effective after original accepted; effective before original rejected without mutating the entity.
- Integration (`transaction_void_with_effective`): posts on an effective date, asserts the too-early void is rejected, then voids on the original date and asserts cumulative effective balances (BTC + USD, settled + pending) net to zero on that date.
- Full `cala-ledger` suite passes (85/85); event schemas unchanged.

### DIFF
--- a/cala-ledger/src/ledger/mod.rs
+++ b/cala-ledger/src/ledger/mod.rs
@@ -219,15 +219,20 @@ impl CalaLedger {
         Ok(transaction)
     }
 
+    /// Voids a transaction by posting a new transaction with reversed entries on the
+    /// given effective date — typically the original transaction's effective date while
+    /// the accounting period is still open, or the current date once it is closed. The
+    /// date may not precede the original transaction's effective date.
     #[instrument(name = "cala_ledger.void_transaction", skip(self))]
     pub async fn void_transaction(
         &self,
         voiding_tx_id: TransactionId,
         existing_tx_id: TransactionId,
+        effective: chrono::NaiveDate,
     ) -> Result<Transaction, LedgerError> {
         let mut db = self.begin_operation().await?;
         let transaction = self
-            .void_transaction_in_op(&mut db, voiding_tx_id, existing_tx_id)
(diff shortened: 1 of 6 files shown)


nicolasburtey:
Superseded by #755: rather than fixing void's effective-date semantics, we're removing the voiding capability — it's the only posting path that bypasses the tx-template system, and no consumer uses it (lana-bank does corrections as domain-level compensating transactions). If voiding is ever reintroduced, it should take an explicit effective date from day one; this PR has that design.

Issue #746, opened 2026-07-15
account_set: global advisory lock (ADDVISORY_LOCK_ID = 123456) serializes all membership changes ledger-wide


nicolasburtey:
## Summary

`account_set/repo.rs` takes a single **global** advisory transaction lock for every account-set membership mutation:

```rust
const ADDVISORY_LOCK_ID: i64 = 123456;   // src/account_set/repo.rs:15
...
sqlx::query!("SELECT pg_advisory_xact_lock($1)", ADDVISORY_LOCK_ID)
```

All four membership operations acquire it (`add_member_account_and_return_parents`, `remove_member_account_and_return_parents`, `add_member_set_and_return_parents`, `remove_member_set_and_return_parents` — lines 370/445/517/602 on `main` as of 2026-07-15).

Two properties combine into a ledger-wide serialization point:

1. **The key is a constant.** Membership changes on completely disjoint set hierarchies contend with each other.
2. **It is xact-scoped**, so it is held until the *enclosing* transaction commits. Callers that batch membership adds into larger transactions (via `add_member_in_op`) hold the global lock through everything else that transaction does — account creation, transaction posting, balance updates.

Under concurrent load, throughput of *any* code path that touches account-set membership collapses to one-transaction-at-a-time.

## Evidence (lana-bank staging, cala-ledger 0.15.12)

Environment: `sb-realtime-high` (high-volume realtime simulation), Cloud SQL Postgres. Load profile: ~11 credit-facility activations + ~11 customer onboardings per minute; each facility activation creates ~10 accounts and adds each to a summary set inside one job transaction (~250–400 membership adds/min).

- Cloud SQL Query Insights: `SELECT pg_advisory_xact_lock($1)` was the **top query by total execution time** — 1,126 calls, **451 ms avg**, in a 3.5-minute window (2026-07-15 15:44–15:47Z). The average mixes instant same-transaction re-acquisitions with multi-second first-acquire waits, so the tail is much worse.
- Live `pg_locks`: **1 holder + 7 waiters** queued on `(classid=0, objid=123456)`, sustained over minutes.
- The sampled holder kept the lock **≥3 s** and simultaneously held ~20 `(classid=1, hashtext(account_id))` balance locks — i.e. it was mid-way through balance posting while everyone else queued behind the membership lock it had acquired earlier in the transaction.
[...]

PR #748, opened 2026-07-15
fix(account_set): stop serializing account-member ops on the global membership lock

nicolasburtey:
Addresses #746 — suggestion 2, investigated. This is the **minimal alternative** to the per-hierarchy locking draft #747; the two are alternatives, not stackable as-is.

## Problem

Every account-set membership mutation takes the single global advisory lock `ADDVISORY_LOCK_ID = 123456` in exclusive mode and — being xact-scoped — holds it until the *enclosing* transaction commits. Callers like lana-bank batch ~10 membership adds inside large transactions, so all membership-touching transactions ledger-wide serialize. On lana's staging sim this was the top query by total DB time (451 ms avg wait, queue of 1 holder + 7 waiters, worst span 32.8 s of pure lock wait). Full evidence in #746.

## What we learned about "shorten the hold time"

#746's second suggestion was to shrink the hold window (acquire later / isolate the closure maintenance). That turns out to be structurally blocked:

- **The xact-scope is load-bearing.** Each mutation is a read-then-write over the ancestor chain; the walk's snapshot must stay valid until the writes commit, otherwise a concurrent account-add and set-attach each miss the other's uncommitted rows (write-skew) and the parent chain silently loses closure rows. The lock cannot be released before commit.
- **Deferring the mutation to commit time** (es-entity commit hooks would support it) breaks same-transaction posting: entry fan-out to parent sets reads the closure table, and callers add an account to its sets and post to it in the same transaction. The closure rows must exist before the post.

So the implementable version of "make the convoy window small" is: keep the coarse key, but stop hot-path holders from excluding *each other*.

## Change

Split the protocol by mutation type (~15 lines in `account_set/repo.rs`):

- **Set-structure mutations** (`add/remove_member_set`): unchanged — EXCLUSIVE coarse lock. They mutate the edges every walk reads and read the member rows account-member mutations write, so they must exclude everything.
- **Account-member mutations** (`add/remove_member_account`): coarse lock **SHARED** + an **EXCLUSIVE per-member advisory lock** (2-arg, `classid 2`, `hashtext(member_account_id)`). Account-member mutations only read the edges, and closure writes for different members are disjoint rows, so they can safely run concurrently. The per-member lock serializes add/remove of the *same* member in overlapping hierarchies, whose interleaved inserts/deletes on shared ancestors would otherwise tear the closure.

The hot path (lana: hundreds of account adds per minute from facility/customer creation) stops convoying; only genuine structure changes (chart edits, product provisioning) still serialize globally.

Acquisition order is fixed (coarse first, then member — two statements): waiting on the coarse lock while holding a member lock could form a wait cycle with a queued exclusive waiter under PostgreSQL's FIFO lock queueing.

## Why this is deploy-safe and low-risk

- **Rolling deploys stay safe**: old binaries take the coarse lock exclusive for account-member ops, which correctly (if conservatively) serializes against both modes used by new binaries. (The per-hierarchy PR does not have this property — this PR can also serve as its safe intermediate version.)
- **No new deadlock surface** for the realistic workload: member locks only conflict on the same member id; structure ops take no member locks; re-granting an already-held shared lock does not queue.
- Same-member duplicate adds still surface as `MemberAlreadyAdded` via the existing unique constraint, exactly as before.

## Testing

- New `tests/account_set_membership_locks.rs`:
  - adds for different members don't serialize (fails on `main` by lock-wait timeout);
  - same-member ops still serialize; structure ops still fence account-member ops (regression guards);
  - write-skew race guard: 16 concurrent adds into a 3-deep hierarchy, closure completeness asserted from the table.
  - Each test creates a throwaway database: these tests hold advisory locks open on purpose, and advisory locks are per-database, so sharing `PG_CON`'s database with other suites would cause FIFO-queue interference.
- Full `cala-ledger` suite: 86/86 pass (including `ec_recalc_race` and `add_member_multi_call_no_deadlock_with_posters`).

Draft because we'd like to load-test against the lana-bank staging environment that reproduces the convoy continuously before merging, and because the maintainers should pick between this and the per-hierarchy draft #747.

### DIFF
--- a/cala-ledger/.sqlx/query-089ea282c32176080fdf70f1d1b62d649ddf49e4333effc13729bbeb395190b7.json
+++ b/cala-ledger/.sqlx/query-089ea282c32176080fdf70f1d1b62d649ddf49e4333effc13729bbeb395190b7.json
@@ -1,0 +1,23 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT pg_advisory_xact_lock($1, hashtext($2))",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "pg_advisory_xact_lock",
+        "type_info": "Void"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Int4",
+        "Text"
+      ]
+    },
+    "nullable": [
+      null
+    ]
+  },
+  "hash": "089ea282c32176080fdf70f1d1b62d649ddf49e4333effc13729bbeb395190b7"
+}
--- a/cala-ledger/.sqlx/query-0adaf45623673e3453f08755801f533c15c19b5350e5d6fa1a92dd1486391830.json
+++ b/cala-ledger/.sqlx/query-0adaf45623673e3453f08755801f533c15c19b5350e5d6fa1a92dd1486391830.json
@@ -1,0 +1,22 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT pg_advisory_xact_lock_shared($1)",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "pg_advisory_xact_lock_shared",
+        "type_info": "Void"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Int8"
+      ]
+    },
+    "nullable": [
+      null
+    ]
+  },
+  "hash": "0adaf45623673e3453f08755801f533c15c19b5350e5d6fa1a92dd1486391830"
+}
--- a/cala-ledger/src/account_set/repo.rs
+++ b/cala-ledger/src/account_set/repo.rs
@@ -12,7 +12,67 @@ use crate::{
 
 use super::{entity::*, error::*};
 
+/// Coarse advisory lock guarding the account-set membership graph
+/// (`cala_account_set_member_accounts` and
+/// `cala_account_set_member_account_sets`).
+///
+/// Membership maintenance is a read-then-write over the whole ancestor
+/// chain: each mutation walks the set-to-set edges (the recursive
+/// `parents` CTE) and then writes the transitive-closure rows the walk
+/// justifies. Two concurrent mutations that each miss the other's
+/// uncommitted writes would leave the closure table inconsistent
+/// (write-skew), so the walk's snapshot has to stay valid until the
+/// writes commit — which is why every lock here is transaction-scoped
+/// and held to commit; releasing earlier would reopen the race.
+///
+/// Lock protocol:
+///
+/// - Set-structure mutations (`add_member_set` / `remove_member_set`)
+///   take this lock EXCLUSIVE. They mutate the edges that every walk
+///   reads, and read the member rows that account-member mutations
+///   write, so they must exclude everything.
+/// - Account-member mutations (`add_member_account` /
+///   `remove_member_account`) take this lock SHARED plus an EXCLUSIVE
+///   per-member lock (`MEMBER_LOCK_CLASS`, keyed on the member account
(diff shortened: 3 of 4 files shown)


nicolasburtey:
Local before/after load-test results for this branch (via the lana-bank pin in GaloyMoney/lana-bank#7289): at 12-way parallel loan creation the membership lock drops from **0.78 ms mean / 64 ms max / 1,693 ms total wait (top advisory query)** to **0.00 ms mean / sub-ms max / 5 ms total**, with identical loan throughput, pipeline latency, and balance-lock profile — full table in https://github.com/GaloyMoney/lana-bank/pull/7289#issuecomment (results comment). At the exact staging cadence (loan every 2 s) the two builds are behaviorally identical — no regression. Staging validation on sb-realtime-high is the remaining step.

nicolasburtey:
Corrected local benchmark (setup-free: shared product, 3 runs/side, min-of-3, details in GaloyMoney/lana-bank#7289): wall time 70.1s → 68.3s (−2.6%) with run-to-run spread collapsing from 19% to 0.9%; membership-lock wait 1,765ms total / 87ms max / queue in 5.5% of samples → 5ms total / sub-ms max / zero queue. 360/360 loans succeeded on each side.

nicolasburtey:
## Local rate-ladder benchmark — the convoy reproduced and eliminated

Validation of this branch through lana-bank (pin: GaloyMoney/lana-bank#7289). Method: full local stack, setup-free loan pipeline (shared credit-facility product; per loan: prospect → customer → deposit account → proposal → approval → activation → disbursal ≈ 18 membership-lock acquisitions), paced arrivals over 3-minute windows, fresh DB per side, 3 runs per side, min-of-3 selected, lock waits from `pg_stat_statements`.

**Ladder (p50 latency of min run, BEFORE 0.15.12 → AFTER this branch):**

| Arrival rate | p50 | Membership-lock aggregate wait / run |
|---|---|---|
| 1 loan/s (180) | 2.48 s → 2.48 s | 37 ms → <1 ms |
| 2 loans/s (360) | 2.47 s → 2.49 s | ~230 ms → ~27 ms |
| **4 loans/s (720)** | **16.58 s → 11.64 s (−30 %)** | **36.6 s → 0.14 s (−99.6 %)** |

**At the 4 loans/s knee (min-of-3):**

| | BEFORE | AFTER |
|---|---|---|
| Wall time | 203.6 s | 195.8 s |
| Latency p50 / p95 / max | 16.58 / 38.58 / 43.71 s | 11.64 / 23.70 / 29.98 s |
| Membership-lock wait: aggregate / worst run / max single | 36.6 s / 73.2 s / 484 ms | 0.14 s / 0.07 s* / <1 ms |
| Balance-lock (classid 1) aggregate | 1.7 s (worst 11.6 s) | 2.4 s (worst 19.2 s) |

\* worst-run value is lower because run 1 was the heaviest AFTER run for balance locks, not membership.

Reading:

- Below the knee the builds are identical (regression check at the production controller cadence: clean).
- Past the knee, the old global lock consumes 20–40 % of aggregate worker time (single waits up to 484 ms — approaching the 451 ms *mean* observed on the staging environment in #746), and this branch converts that into a 30–40 % p50/p95 latency reduction with the membership lock dropping out of the advisory-lock profile entirely.
- With the membership convoy gone, contention moves to the classid-1 **balance** locks (all facilities in the test share one product's non-EC summary/omnibus accounts, so per-balance exclusive locks serialize posting) — expected to be the next hotspot in ledgers with hot shared accounts; independent of this PR.

Remaining validation: deploy against the staging environment that reproduces the convoy continuously (tracked on the lana side).

bodymindarts (approved):

PR #782, opened 2026-07-28
perf(ledger): batch member attach + drop dead parent-list returns

nicolasburtey:
## Why

Profiling a 10-loans/s lana stress test (30-min soak, custom release build on a staging sandbox), the single hottest statement was the `WITH RECURSIVE parents` ancestor walk in `add_member_account_and_return_parents`:

- **~0.5–0.6 DB-cores steady state** (65–80 calls/s at ~4 loans/s achieved), mean 8–10ms
- 1:1 with account creation — lana creates **~19 ledger accounts per loan** (`create_accounts_for_credit_facility_in_op`), each attached to account sets one-by-one → one recursive walk per attach
- cumulative 31M tuples seq-read on `cala_account_sets` (2k rows) and 21M on `cala_account_set_member_account_sets` in 13 minutes

Two findings, two commits:

## 1. `refactor`: the walk's return value is dead

All four `add/remove_member_{account,set}_and_return_parents` functions computed `(DateTime, Vec<AccountSetId>)` via an outer `SELECT ... FROM parents UNION ALL SELECT NULL, NOW()` — and **every caller discards it**. The statements now end in the data-modifying CTE itself and return `()`. Also drops the redundant `cala_account_sets` join in the walk's anchor term (the FK on `account_set_id` already guarantees the parent set row exists). Public API (`add_member_in_op` etc.) unchanged.

## 2. `feat`: batch member attach — one walk for N pairs

New `AccountSets::add_members` / `add_members_in_op` taking `&[(AccountSetId, AccountId)]`. Single-attach semantics preserved:

- target sets resolved in one `find_all_in_op`
- the no-balance-history guard runs as **one lock query + one existence query** (new `Balances::members_with_balance_history_in_op`) instead of 2N queries
- all memberships (incl. transitive ancestor rows) inserted by **one recursive statement** covering every pair; per-member advisory locks taken in one id-ordered statement (same lock protocol as the single path)
- one `AccountSetMemberCreated` outbox event per pair
- duplicate pairs / members with balance history error exactly as the single-attach path does

For the lana hot path this collapses ~19 walks per loan into 1.

## Follow-up (separate PR, lana-bank)

Adopt in `core/credit/src/ledger/mod.rs`: collect the per-facility `(account_set_id, account_id)` pairs from `create_accounts_for_credit_facility{,_proposal}_in_op` and call `add_members_in_op` once per facility instead of `add_member_in_op` per account. Requires a cala release containing this PR.

## Tests

- new `add_members_batch` integration test: rollup equivalence with the single-attach path, empty-batch no-op, duplicate-member error, unknown-set error
- full suite: **95/95 pass**

## Verification plan (post-release)

Re-run the lana stress sandbox (10 loans/s, `stress-testing` preset) with the cala bump and diff queryid `WITH RECURSIVE parents`: expect calls/s to drop ~19× and its share of DB time to fall out of the top ranks.


nicolasburtey:
Superseded by the split: #783 (refactor: drop dead parent-list returns) + #784 (feat: batch member attach, stacked on #783).

PR #798, opened 2026-08-04
feat(ledger): journal entry listing filtered by created-at + effective date (Option A for lana #7796)

bodymindarts:
## Summary

Adds a typed cala query method for **journal-scoped entry listing filtered by
created-at range and by the posting transaction's effective date**, cursor-paginated
on `(created_at, id)` in either direction.

This is the exact shape es-entity's generated `list_for_*` methods can't express:
they do equality on a single column + created-at cursor only, so a **created-at
range window** and the **cross-entity effective-date filter** (`effective` lives on
`cala_transactions`, not `cala_entries`) require a hand-written query.

```rust
pub struct EntriesFilter {
    pub created_at_from: Option<DateTime<Utc>>,
    pub created_at_to: Option<DateTime<Utc>>,
    pub effective_from: Option<NaiveDate>,
    pub effective_to: Option<NaiveDate>,
}

// on Entries
pub async fn list_for_journal_id_filtered(
    &self,
    journal_id: JournalId,
    filter: EntriesFilter,
    query: PaginatedQueryArgs<EntryByCreatedAtCursor>,
    direction: ListDirection,
) -> Result<PaginatedQueryRet<Entry, EntryByCreatedAtCursor>, EntryError>
```

## What's in the change

- **`Entries::list_for_journal_id_filtered`** + `EntriesFilter` (`cala-ledger/src/entry/`).
  Implemented as a hand-written `es_query!` mirroring the existing
  `list_for_account_set_id_by_created_at` precedent — all bounds inclusive, AND
  semantics, unset bounds ignored.
- **Two indexes added to the `cala_ledger_setup` migration in place** (a breaking
  migration edit — acceptable pre-production; recreate dev stacks):
  - `idx_cala_entries_journal_id_created_at (journal_id, created_at DESC, id DESC)` —
    this backs the **existing** `Entries::list_for_journal_id` too, which was
    previously **unindexed** (scan + sort). Standalone perf win independent of the
    filter.
  - `idx_cala_transactions_effective (effective)` — backs the effective-date range.
- **Regenerated `.sqlx` offline cache** (2 new query files).
- **Integration test** `tests/entries_filter.rs` covering effective range, created-at
  range, no-filter, AND composition, empty result, and cursor pagination composed
  with a filter.

## Design notes

The outer query stays **single-table on `cala_entries`** (so `created_at`/`id` are
unambiguous and pagination matches the other `list_for_*` methods); the effective-date
range is applied as a **semi-join** against `cala_transactions`
(`transaction_id IN (SELECT id FROM cala_transactions WHERE effective …)`). When
neither effective bound is set the disjunct short-circuits to the plain journal
listing, so the unfiltered path keeps its fast plan.

No new events, no change to the outbox surface, no dependency changes.

## Why — relationship to lana-bank

lana-bank [#7796](https://github.com/GaloyMoney/lana-bank/pull/7796) ("journal
date-range filtering POC") implements this capability by querying cala's tables
directly — a `CREATE VIEW` over `cala_entries ⋈ cala_entry_events (JSONB) ⋈
cala_transactions ⋈ cala_accounts` plus two indexes on cala's tables, all from
**lana's** migration, then raw `sqlx` run on cala's pool. That couples lana to cala's
physical schema, event-sourcing JSONB layout, enum serde casing, and an internal
`sequence = 1` invariant.

This PR is **Option A** from the cross-DB boundary analysis (drua `cala-dev`
space, `analysis-lana-pr7796-cross-db-query.md`): expose the capability as a typed
cala API so lana can drop the cross-schema coupling.

**Migration path for #7796 once this lands:** `Journal::filtered_entries` collapses to
`self.cala.entries().list_for_journal_id_filtered(journal_id, filter.into(), cursor, Descending)`
mapping `Entry -> LedgerEntry` via the existing `TryFrom`; lana deletes its
`core_journal_entries` view, the two `CREATE INDEX`es on cala tables, the raw
`QueryBuilder` SQL, and the enum re-parsing. The GraphQL + admin-panel layers are
unchanged.

## Validation

- `nix flake check` — green (fmt, clippy `--all-features --deny warnings`, audit, deny).
- `cargo sqlx prepare` against a migrated DB — new queries type-check; cache committed.
- `cargo nextest run --test entries_filter` — passes.

## Status

Draft / POC — opening for design feedback on the API shape (filter struct fields,
method naming, whether to fold the follow-up filters — entry_type, layer, account-code
scoping, sort — into `EntriesFilter` now vs. incrementally).

---

> [!NOTE]
> **Medium Risk**
> Touches ledger read paths and alters the base migration (index additions); query correctness and pagination semantics matter for downstream consumers like lana-bank, but behavior is covered by integration tests and does not change posting or event surfaces.
> 
> **Overview**
> Adds **`EntriesFilter`** and **`Entries::list_for_journal_id_filtered`**, a typed API to list journal entries with optional inclusive bounds on entry `created_at` and on the parent transaction’s **effective** date, with cursor pagination on `(created_at, id)` in ascending or descending order. Generated `list_for_*` helpers cannot express created-at ranges or filters on `cala_transactions.effective`, so this uses a hand-written `es_query!` that keeps the outer query on `cala_entries` and applies effective-date constraints via a semi-join on `cala_transactions`.
> 
> The setup migration gains **`idx_cala_entries_journal_id_created_at`** (also benefits existing `list_for_journal_id`) and **`idx_cala_transactions_effective`**. Offline **`.sqlx`** cache is updated; integration test **`entries_filter`** covers filters, composition, empty results, and pagination.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 59d09d23901dc7cbc264977203c4b9084cddb642. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cala-ledger/.sqlx/query-3d6e1d30a0040d09c2a5e1fcd84bb17aa3dcba84905d9fcc0b2749efabbaa0d9.json
+++ b/cala-ledger/.sqlx/query-3d6e1d30a0040d09c2a5e1fcd84bb17aa3dcba84905d9fcc0b2749efabbaa0d9.json
@@ -1,0 +1,60 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "WITH entities AS (\n                    SELECT created_at, id\n                    FROM cala_entries\n                    WHERE journal_id = $4\n                      AND ($5::timestamptz IS NULL OR created_at >= $5::timestamptz)\n                      AND ($6::timestamptz IS NULL OR created_at <= $6::timestamptz)\n                      AND (\n                        ($7::date IS NULL AND $8::date IS NULL)\n                        OR transaction_id IN (\n                          SELECT id FROM cala_transactions\n                          WHERE ($7::date IS NULL OR effective >= $7::date)\n                            AND ($8::date IS NULL OR effective <= $8::date)\n                        )\n                      )\n                      AND (COALESCE((created_at, id) < ($3, $2), $2 IS NULL))\n                    ORDER BY created_at DESC, id DESC\n                    LIMIT $1) SELECT i.id AS \"entity_id: Repo__Id\", e.sequence, e.event, CASE WHEN $9 THEN e.context ELSE NULL::jsonb END as \"context: es_entity::ContextData\", e.recorded_at, NULL::jsonb as \"forgettable_payload?\" FROM entities i JOIN cala_entry_events e ON i.id = e.id ORDER BY i.created_at desc, i.id desc, i.id, e.sequence",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "entity_id: Repo__Id",
+        "type_info": "Uuid"
+      },
+      {
+        "ordinal": 1,
+        "name": "sequence",
+        "type_info": "Int4"
+      },
+      {
+        "ordinal": 2,
+        "name": "event",
+        "type_info": "Jsonb"
+      },
+      {
+        "ordinal": 3,
+        "name": "context: es_entity::ContextData",
+        "type_info": "Jsonb"
+      },
+      {
+        "ordinal": 4,
+        "name": "recorded_at",
+        "type_info": "Timestamptz"
+      },
+      {
+        "ordinal": 5,
+        "name": "forgettable_payload?",
+        "type_info": "Jsonb"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Int8",
+        "Uuid",
+        "Timestamptz",
+        "Uuid",
+        "Timestamptz",
+        "Timestamptz",
+        "Date",
+        "Date",
+        "Bool"
+      ]
+    },
+    "nullable": [
+      false,
+      false,
+      false,
+      null,
+      false,
+      null
+    ]
+  },
+  "hash": "3d6e1d30a0040d09c2a5e1fcd84bb17aa3dcba84905d9fcc0b2749efabbaa0d9"
+}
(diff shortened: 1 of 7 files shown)


nicolasburtey (commented):
Reviewed the full diff — this is the right upstream shape and supersedes the cross-schema coupling in lana #7796 cleanly. Specifically like: single-table outer query + semi-join for `effective` (cursor semantics identical to other `list_for_*`, no NULL pitfalls), inclusive bounds matching lana's GraphQL contract 1:1, `first+1`/`fetch_n` pagination convention, and the journal index as a standalone win for the previously unindexed `list_for_journal_id`.

Notes from the review:

1. ~~Index-name collision with lana #7796~~ — moot per merge order (cala first, lana rebases and drops its duplicates).
2. **Ascending direction was untested** — the ASC query is a separate SQL string (`>` predicate, `ASC` order) that could drift silently. Fixed in 59d09d23: adds ASC ordering assertions on `(created_at, id)` (including the id tie-break for same-timestamp entries posted in one tx), gap/overlap-free cursor continuation, and ASC composed with an effective filter. Test passes locally.
3. Non-sargable catch-all predicates (`($5 IS NULL OR created_at >= $5)`) — leaving as-is: PG still drives from the new `journal_id` composite index, and the strict fix (static sargable variants) belongs in es-entity range support, not a hand-written query.
4. ASC/DESC predicate duplication — consistent with the generated code; the new ASC coverage is the drift guard.

LGTM to land when ready; lana #7796 will rebase onto the released version and collapse `Journal::filtered_entries` to this call.

PR #799, opened 2026-08-04
fix(account-set): reject self-membership that double-counts balances

nicolasburtey:
## Problem

`AccountSets::add_member(_in_op)` / `add_members_in_op` accepted adding an account set's **own underlying account** as a member (account sets *are* accounts, and `AccountId::from(AccountSetId)` is public). The posting path resolves an entry's account to its member sets **and** the account itself:

```rust
mappings.get(&entry.account_id)      // -> [S]  (== entry.account_id!)
    .chain(std::iter::once(entry.account_id))  // -> [S, S]
```

so every entry was applied **twice** to the same balance — in `Balances::new_snapshots`, in the effective-balance path, and in velocity `balances_to_check` — silently inflating amounts with no error raised.

## Fix

- Reject self-membership (both the `Account` and `AccountSet` member forms, single and batch paths) with a new `AccountSetError::CannotAddSelfAsMember`.
- Defense in depth: filter self-mappings out of the account-resolution chains in `Balances::new_snapshots`, `EffectiveBalances::update_cumulative_balances_in_op` and `VelocityBalances::balances_to_check`, so even a pre-existing/crafted self-mapping can no longer double-apply an entry.

## Tests

- New unit test: `new_snapshots_applies_entry_once_for_self_mapping`
- New integration test: `errors_on_self_membership` (single account form, set form, and batch path)
- Full workspace suite passes (109 tests)

### DIFF
--- a/cala-ledger/src/account_set/error.rs
+++ b/cala-ledger/src/account_set/error.rs
@@ -48,6 +48,12 @@ pub enum AccountSetError {
          only eventually-consistent sets support recalculation"
     )]
     CannotRecalculateNonEcSet { account_set_id: AccountSetId },
+    #[error(
+        "AccountSetError - Cannot add account set '{account_set_id}' as a member of itself: \
+         self-membership would make every entry to the underlying account count twice \
+         in the set's balance"
+    )]
+    CannotAddSelfAsMember { account_set_id: AccountSetId },
 }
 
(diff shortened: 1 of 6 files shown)


nicolasburtey:
cc @bodymindarts

PR #800, opened 2026-08-04
fix(account-set): prevent membership cycles in add_member_set

nicolasburtey:
## Problem

`add_member_set` inserted set→set edges with no cycle check. Creating A→B and then B→A has two severe consequences:

1. **DoS**: every membership mutation walks ancestors via `WITH RECURSIVE parents ... UNION ALL` with no cycle detection. Once a cycle exists, any subsequent add/remove touching it never terminates — while holding the coarse `ADDVISORY_LOCK_ID` **exclusive** advisory lock, blocking *all* membership mutations ledger-wide.
2. **Closure corruption**: the transitive member closure cross-contaminates both sets (each set's members folded into the other), so balance rollups silently include each other's members.

## Fix

Reject the edge up front with `AccountSetError::MembershipCycleDetected` when the candidate member is the target itself or is found in the target's ancestor walk. The check runs under the same `EXCLUSIVE` coarse advisory lock that guards the insert, so the validated graph cannot change before commit.

Note: the direct self-edge (`add_member(S, S)`) is already rejected earlier with `CouldNotFindById` by the entity resolution in `add_member_in_op`; the repo-level self-check added here makes the guard explicit regardless of call path. (See also #799, which gives the self case a dedicated error.)

## Tests

- New integration test `errors_on_membership_cycle`: direct cycle (A↔B), transitive cycle (A→B→C then C→A), self-edge, and a non-cyclic control case.
- New query added to `.sqlx` via `cargo sqlx prepare` (only the new query file; the 35 unrelated deletions a local `prepare` produces on clean `main` were reverted — that's a pre-existing environment discrepancy).


nicolasburtey:
cc @bodymindarts

PR #801, opened 2026-08-04
fix(tx-template): reject non-positive entry units

nicolasburtey:
## Problem

Nothing validated that a template's `units` expression evaluates to a positive amount. A `DEBIT -X` is accounting-equivalent to `CREDIT +X`, but everything downstream treats the entry by its direction label:

- The zero-sum double-entry check in `prep_entries` still passes (it nets signed values).
- `Snapshots::update_snapshot` adds the negative units to the dr/cr legs, driving individual legs negative.
- Velocity `enforce()` checks net dr−cr against the limit, so negative net amounts always pass — a transfer executed entirely with negative-unit entries inverts each side's direction and never trips debit/credit velocity windows.

Any template using e.g. `units: params.amount` was one negative param away from bypassing velocity controls.

## Fix

Reject `units <= 0` in `prep_entries` with a new `TxTemplateError::NonPositiveUnits` that names the offending entry type and the evaluated amount.

## Tests

- New integration test `errors_on_non_positive_units`: negative and zero amounts are rejected, positive amount posts fine.

### DIFF
--- a/cala-ledger/src/tx_template/error.rs
+++ b/cala-ledger/src/tx_template/error.rs
@@ -33,6 +33,10 @@ pub enum TxTemplateError {
     SerdeJson(#[from] serde_json::Error),
     #[error("TxTemplateError - UnbalancedTransaction: currency {0}, layer {1:?}, amount {2}")]
     UnbalancedTransaction(Currency, Layer, Decimal),
+    #[error(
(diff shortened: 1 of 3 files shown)


nicolasburtey:
cc @bodymindarts

PR #802, opened 2026-08-04
fix(ledger): reject entries targeting eventually-consistent accounts

nicolasburtey:
## Problem

The posting path deliberately skips eventually-consistent (EC) accounts when updating balances — `BalanceRepo::find_for_update` filters them out of both the lock query and the data fetch, and EC account sets are maintained by recalculation from *member* history. But nothing stopped a template from posting an entry **directly to an EC set's underlying account**:

- the entry was persisted to `cala_entries` and published to the outbox,
- no balance-history row was ever written for it (`new_snapshots` hits `current = None → continue`),
- and no recalc would ever see it (recalc only folds in *member* history).

The value simply disappeared from all balances while remaining in the entry log — a silent ledger/balance divergence.

## Fix

Reject such postings in `post_transaction_in_op` with a new `LedgerError::EntriesTargetEventuallyConsistentAccount`, checked right after entries are prepared (before velocity and balance updates) so the whole operation rolls back.

## Tests

- New integration test `errors_when_entry_targets_eventually_consistent_account`: posting to an EC set's account errors; posting the same template to regular accounts succeeds.

### DIFF
--- a/cala-ledger/src/ledger/error.rs
+++ b/cala-ledger/src/ledger/error.rs
@@ -34,6 +34,11 @@ pub enum LedgerError {
     BalanceError(#[from] BalanceError),
     #[error("LedgerError - VelocityError: {0}")]
     VelocityError(#[from] VelocityError),
+    #[error(
+        "LedgerError - EntriesTargetEventuallyConsistentAccount: entries cannot target \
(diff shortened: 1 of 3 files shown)


nicolasburtey:
cc @bodymindarts

PR #803, opened 2026-08-04
fix(velocity): record balance-key account id in new velocity snapshots

nicolasburtey:
## Problem

When a velocity control is attached to an **account set**, the first snapshot for a `(set, window)` balance key was built with the *member* entry's `account_id`:

```rust
None => crate::balance::Snapshots::new_snapshot(time, entry.account_id, entry),
```

The DB columns were correct (they come from `VelocityBalanceKey`), but the persisted JSON `values.account_id` was wrong — and stayed wrong forever, since `update_snapshot` propagates the existing `account_id`. That misidentifies the account in `LimitExceededError` and for any consumer reading velocity balance values.

## Fix

Pass `key.account_id` to `Snapshots::new_snapshot`, matching what `Balances::new_snapshots` already does on the main balance path.

## Tests

- New unit test `new_snapshots_uses_key_account_id_not_entry_account_id` covering the account-set case (key account ≠ entry account).

### DIFF
--- a/cala-ledger/src/velocity/balance/mod.rs
+++ b/cala-ledger/src/velocity/balance/mod.rs
@@ -147,7 +147,10 @@ impl VelocityBalances {
                     Some(balance) => {
                         crate::balance::Snapshots::update_snapshot(time, balance, entry)
                     }
-                    None => crate::balance::Snapshots::new_snapshot(time, entry.account_id, entry),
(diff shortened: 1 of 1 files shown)


nicolasburtey:
cc @bodymindarts

bodymindarts (approved):

PR #805, opened 2026-08-04
docs: document advisory-lock ordering invariants

nicolasburtey:
Documentation-only PR capturing the conclusions of the locking review (finding 7), so the invariants survive future edits. **No behavior change.**

## What the review found

The advisory-lock protocol itself held up: canonical ordering is consistent across posters, EC recalcs, and both member-check paths; the 2-arg EC-set namespace and the membership coarse/per-member locks don't produce a deadlock cycle I could construct. But two fragile points were only implicitly protected:

1. **`find_for_update` (balance and velocity) relies on caller-side input sorting, not the query's `ORDER BY`.** The code already notes elsewhere (`lock_accounts_exclusive_in_op`) that PostgreSQL's planner is free to evaluate projection expressions (the `pg_advisory_xact_lock` calls) *before* any sort node. The canonical acquisition order is safe today only because callers pre-sort the UNNEST input (BTreeSet / `sorted_keys`) and that order coincides with the `ORDER BY`. The doc comments now say this explicitly and warn against reordering either side.
2. **All 1-arg advisory locks share one namespace** (balance locks, velocity locks, `ADDVISORY_LOCK_ID = 123456`), and all keys are 32-bit `hashtext` values. Collisions cause false serialization only — documented at `ADDVISORY_LOCK_ID` and both `find_for_update` sites. Also documented that the velocity lock key deliberately excludes `window` (safe over-serialization).

### DIFF
--- a/cala-ledger/src/account_set/repo.rs
+++ b/cala-ledger/src/account_set/repo.rs
@@ -45,6 +45,16 @@ use super::{entity::*, error::*};
 /// lock. An operation must never wait on the coarse lock while holding
 /// a per-member lock — under PostgreSQL's FIFO lock queueing that can
 /// form a wait cycle with a queued exclusive (structure) waiter.
+///
+/// Namespace note: this is a 1-arg `pg_advisory_xact_lock`, so it
+/// shares a key space with the balance locks
+/// (`hashtext(journal_id, account_id, currency)`), the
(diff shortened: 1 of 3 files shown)


nicolasburtey:
cc @bodymindarts

bodymindarts:
Outdated and false.

PR #811, opened 2026-08-05
feat(ledger): stream EC account-set balance rollup from the outbox

bodymindarts:
Replaces the pull/batch eventually-consistent (EC) account-set balance recalc with a streaming job that consumes the obix outbox in sequence order and folds each committed transaction's leaf deltas into its ancestor EC sets, committing in bounded batches.

Supersedes #753 (same feature, reimplemented on the obix 0.6 primitives after #106/#95/#86 landed).

## Implementation

- The rollup is an obix `OutboxEventHandler` with a persistent-only subscription: `TransactionCreated` events are **collect**ed into the pending batch (a pure memory write — no transaction per event), everything else is skipped. One `flush` per batch landing applies all rollups **inside the transaction that commits the checkpoint** — rollup writes and stream position land atomically (exactly-once; the applier is additive, not idempotent).
- Batching (`max_batch_size = 1000`), checkpointing, replay and shutdown belong to obix's managed runner; registration via `Outbox::register_event_handler` (single cluster-wide instance via `spawn_unique`).
- The manual `recalculate_balances`/`_batch`/`_deep` engine and its batch member-history replay are removed — the streaming job is the sole EC-set maintainer. Current **and** effective (incl. back-dated) balances are rolled up.
- Preserved guardrails: the `MemberHasBalanceHistory` membership guard (a member can only join/leave an EC set while it has no balance history) and the poster shared-lock discipline (`EC_SET_LOCK_CLASS`), aligned with the #810 lock-ordering doctrine.
- `CalaLedger::init(config, Option<&mut job::Jobs>)`: pass `Some(&mut jobs)` to register the rollup on a `Jobs` you own and poll; pass `None` and the ledger creates and holds its own (dormant until polled).

## Tests

- `ec_streaming_rollup.rs`: backlog catch-up, nested EC ancestors, EC-vs-inline history equivalence, shared members fanning into multiple sets, effective balances — each on an isolated database (the rollup is a global outbox consumer), in backlog mode (post → `start_poll` → wait for convergence).
- `effective_balance.rs`: EC cumulative effective balances incl. back-dated posts, verified against an inline set.
- Former recalc-driven EC tests in `balance.rs`/`effective_balance.rs` converted to drive the streaming rollup.

---

> [!NOTE]
> **High Risk**
> Changes how EC account-set balances are maintained (core ledger correctness) and alters balance tables/migrations; requires the outbox job to run and updates `CalaLedger::init` to require `Jobs`.
> 
> **Overview**
> Replaces pull/batch EC account-set balance **recalc** (`recalculate_balances` / deep walks, watermark/`latest_seq` replay) with a **streaming outbox job** that incrementally folds each committed transaction into ancestor EC sets. Work per commit stays bounded instead of replaying full member history in one transaction.
> 
> **New `ec_rollup` module:** obix `OutboxEventHandler` on `TransactionCreated` + `EntryCreated`, batches up to 1000 events, flushes rollup writes in the same transaction as the stream checkpoint. Incomplete entry groups in a batch are topped up via `entries.find_all_in_op`. Registered at `CalaLedger::init` with caller-owned `job::Jobs` (re-exported from the crate).
> 
> **Balance layer:** `apply_ec_rollup_in_op` drives settled snapshots (`Snapshots::from_ec_entries`, `fetch_ec_set_mappings`, `find_ec_balances_for_update`) and per-txn effective rollup (`apply_ec_rollup_in_op` on effective balances). Inline posting still uses `Snapshots::from_entries`. **Schema:** drops `cala_balance_history.seq`, `latest_seq`, and the balance-history sequence.
> 
> **Removed:** account-set recalc APIs, EC descendant listing, effective-balance recalc replay SQL, exclusive recalc locks, and `CannotRecalculateNonEcSet`. Membership guard rationale is updated for incremental-from-birth rollup (not watermark races).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4fcb4110595e2007f68ba7cfb2f1df9be322fc1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cala-ledger/src/balance/mod.rs
```diff
@@ -301,224 +306,212 @@ impl Balances {
...
     }
 
-    #[instrument(level = "debug", name = "cala_ledger.balances.new_snapshots", skip_all)]
-    fn new_snapshots(
+    /// Like [`Self::new_snapshots`] but fans each entry **only** into its
+    /// EC ancestor sets (never the leaf account itself, which the inline
+    /// poster path already maintains). Chains repeated writes to the same
+    /// set within the batch so versions increment correctly.
+    #[instrument(name = "cala_ledger.balances.ec_set_snapshots", skip_all)]
+    fn ec_set_snapshots(
```
vindard: The folding in `new_snapshots`/`ec_set_snapshots` isn't really orchestration — it's pure snapshot construction that happens to hang off `Balances`. #812 (stacked on this branch) moves that non-orchestration code out into its own `SnapshotFold` abstraction in `snapshot.rs`, with its tests encapsulated in the same file, leaving `Balances` to just orchestrate. We can either cherry-pick it here, or squash it in if we decide this is a good change

cala-ledger/src/ec_rollup.rs
```diff
@@ -0,0 +1,232 @@
...
+        let missing_ids: Vec<EntryId> = txns
+            .iter()
+            .filter(|tx| entries.get(&tx.id).map_or(0, Vec::len) != tx.entry_ids.len())
+            .flat_map(|tx| tx.entry_ids.iter().copied())
+            .collect();
+        let mut fetched = if missing_ids.is_empty() {
+            HashMap::new()
+        } else {
+            self.entries.find_all_in_op(op, &missing_ids).await?
+        };
```
vindard: Some extra code on the `EcRollupBatch` type that captures whatever entries it sees in the stream, and then backfills what it misses with a `find_all` query. Just double-checking whether we deliberately want this optimization since the tradeoff is that it does make the code less readable

cala-ledger/src/balance/effective/mod.rs
```diff
@@ -193,252 +188,78 @@ impl EffectiveBalances {
...
 
-        let min_effective_date = new_history
-            .iter()
-            .map(|r| r.effective_date)
-            .min()
-            .expect("history is non-empty");
-
-        let set_account_ids: Vec<AccountId> = account_set_ids.iter().map(AccountId::from).collect();
-
+        let new_balances = Self::new_effective_snapshots(journal_id, all_data);
```
vindard: Same shape as the folding above: `new_effective_snapshots` is stateless snapshot construction hanging off `EffectiveBalances` via `Self::`, and it hands back a `(AccountId, Currency, NaiveDate, BalanceSnapshot, u32)` tuple that this call site immediately reassembles into an `EffectiveBalanceSnapshot`. Commit eb875904 (on #812) moves that onto `EffectiveBalanceData::into_snapshots(journal_id)` so the value object builds the snapshot directly and the tuple clump goes away, leaving `EffectiveBalances` doing only orchestration. Could we fold that in alongside the settled-path change?

cala-ledger/src/balance/effective/mod.rs
```diff
@@ -193,252 +188,78 @@ impl EffectiveBalances {
         res
```
vindard: `balance_ranges_from_snapshots` is another stateless transform on the orchestrator — it only builds `BalanceRange` values from their `(open, close)` bounds. Commit a0225514 (on #812) moves that construction onto `BalanceRange::from_bounds` and inlines the collection at the caller, so `find_all_in_range` is left purely orchestration. (The identically-named repo-side helper is a different function and stays untouched.) Could we pull that one in too?

cala-ledger/src/ledger/mod.rs
```diff
@@ -72,6 +88,42 @@ impl CalaLedger {
...
+                    publisher.inner(),
+                    &balances,
+                    &entries,
+                )
+                .await
+                .map_err(LedgerError::EcRollupRegistration)?;
+                None
+            }
+            None => {
+                let mut owned = job::Jobs::init(
```
vindard: This `None` arm has the ledger init and hold its own `Jobs`. Since `cala-ledger` is a library and the owned `Jobs` is never polled here (no accessor, no `start_poll`), should we instead make `jobs: &mut job::Jobs` required and let the caller own `start_poll` (and shutdown)? That matches how `Jobs` was owned before (in the app layer) and drops this branch entirely. Or is there a caller we want the zero-wiring path for?

PR #784, opened 2026-07-28
feat(ledger): batch account-set member attach in a single ancestor walk

nicolasburtey:
> Stacked on #783 (the membership-mutation cleanup) — only the top 2 commits are new; base auto-retargets to `main` once #783 merges.

## Why

Profiling a 10-loans/s lana stress test (30-min soak, custom release build on a staging sandbox), the single hottest statement was the `WITH RECURSIVE parents` ancestor walk in `add_member_account`:

- **~0.5–0.6 DB-cores steady state** (65–80 calls/s at ~4 loans/s achieved), mean 8–10ms
- 1:1 with account creation — lana creates **~19 ledger accounts per loan** (`create_accounts_for_credit_facility_in_op`), each attached to account sets one-by-one → one recursive walk per attach
- cumulative 31M tuples seq-read on `cala_account_sets` (2k rows) and 21M on `cala_account_set_member_account_sets` in 13 minutes

## What changes

New `AccountSets::add_members` / `add_members_in_op` taking `&[(AccountSetId, AccountId)]`. Single-attach semantics preserved:

- target sets resolved in one `find_all_in_op`
- the no-balance-history guard runs as **one lock query + one existence query** (new `Balances::members_with_balance_history_in_op`) instead of 2N queries
- all memberships (incl. transitive ancestor rows) inserted by **one recursive statement** covering every pair; per-member advisory locks taken in one id-ordered statement (same lock protocol as the single path)
- one `AccountSetMemberCreated` outbox event per pair
- duplicate pairs / members with balance history error exactly as the single-attach path does

For the lana hot path this collapses ~19 walks per loan into 1.

## Follow-up (separate PR, lana-bank)

Adopt in `core/credit/src/ledger/mod.rs`: collect the per-facility `(account_set_id, account_id)` pairs from `create_accounts_for_credit_facility{,_proposal}_in_op` and call `add_members_in_op` once per facility instead of `add_member_in_op` per account. Requires a cala release containing this PR.

## Tests

- new `add_members_batch` integration test: rollup equivalence with the single-attach path, empty-batch no-op, duplicate-member error, unknown-set error
- full suite: **95/95 pass**

## Verification plan (post-release)

Re-run the lana stress sandbox (10 loans/s, `stress-testing` preset) with the cala bump and diff queryid `WITH RECURSIVE parents`: expect calls/s to drop ~19× and its share of DB time to fall out of the top ranks.

---

> [!NOTE]
> **Medium Risk**
[...]

PR #814, opened 2026-08-06
feat(account_set)!: adjacency-only membership walk (drops transitive closure, closes #800)

bodymindarts:
## What & why

Deletes the **materialized transitive membership closure**. Membership is now stored as **direct edges only**; an account's ancestor account-sets are resolved at read time by a bounded **recursive upward walk** over the (tiny) set→set edge table.

This implements the `cala-dev` design doc **`impl-pr-walk-only-membership.md`** (PR 3), and is motivated by the holistic locking/perf review (clat memory `019fd1f5`, `cala-dev/analysis-locks-holistic-perf-review.md`): synchronous closure materialization was ~90% of membership write volume (#792: 4.15M closure rows/hour at 10 loans/s) and the recursive ancestor walk was #784's hottest write-path statement. The walk touches only the set→set edge table (~thousands of rows, depth ≤10), measured at ~0.3 ms as #792's live-walk fallback — promoted here to the sole mechanism.

Also **absorbs #800** (membership cycle check + depth cap). **Closes #800.**

> **Base note:** #813 (ec-plain-accounts) is now **merged**, so this PR is rebased directly onto `main` (which contains it). Verified on `main`: `nix flake check` (fmt/clippy `-D`/audit/deny) and `cargo nextest --workspace` both pass.

## Changes

**Schema (fresh-deploy — no in-place migration script)**
- Dropped the `transitive` column from `cala_account_set_member_accounts` by editing the baseline setup migration (`20231208110808`). Both indexes kept. Existing envs must be redeployed fresh; there is intentionally no `ALTER` migration.

**Reads → recursive upward walk**
- `AccountSetRepo::fetch_mappings_in_op` (poster path) and `BalanceRepo::fetch_ec_set_mappings` (EC rollup path) rewritten as `WITH RECURSIVE` walks (seed = direct edges, `UNION` up the set→set edges). `fetch_ec_set_mappings` keeps the `eventually_consistent = TRUE` filter.
- Dropped the now-redundant `transitive IS FALSE` predicate from the three direct member-listing queries.

**Writes → direct-edge only**
- `add_member_account(s)` / `remove_member_account`: single direct insert/delete, closure CTEs removed.
- `add_member_set` / `remove_member_set`: single set→set edge insert/delete, closure propagation removed.

**Double membership stays prohibited (path uniqueness, enforced explicitly)**
The closure's unique constraint was load-bearing: it enforced that an account may be contained in any given set via **at most one membership path** (no double membership — direct-under-a-transitive-ancestor, sibling overlap, diamonds via shared ancestors were all rejected as `MemberAlreadyAdded`). Walk-only keeps that invariant, enforced by explicit read-only validation walks instead of constraint collisions:
- `assert_no_double_membership` (account-member adds, single + batch): expands each new edge's would-be containments (target set + its ancestors) and combines them with the account's existing containments; any set reached twice — via an existing path, the new edge, or across two pairs of the same batch — rejects with the same `MemberAlreadyAdded` error the closure produced.
- `add_member_set` additionally rejects **set-level** double paths (the member already reaches the target or one of its ancestors) and **subtree-account overlap** (an account under the member set already lives under the target chain).
- One deliberate tightening vs. the closure: set-level double paths are rejected **up front even while no accounts are involved**. The old code allowed an empty double edge (e.g. `A⊃B⊃C` plus `A⊃C` — #800's own test asserted this) and then failed every later account-add under it; walk-only rejects the edge itself. (#800's lifted test is adjusted accordingly.)

**Locks — closure-era protocol retained**
The path-uniqueness validation is a read-then-write over the membership graph, so the pre-existing lock protocol survives walk-only unchanged: account-member ops take the coarse lock SHARED plus a per-member EXCLUSIVE lock (`MEMBER_LOCK_CLASS`); set-structure ops take the coarse lock EXCLUSIVE (now guarding cycle check + double-membership checks + depth cap + the single-edge write). Disjoint-member attaches still run concurrently, as post-#748. The perf win of this PR is the **elimination of closure writes** (#792's 4.15M rows/hour and #784's write-path recursive walk), not lock removal.

**#800 (cycle check) + depth cap**
- Lifted #800's cycle-check SQL + `MembershipCycleDetected` error + `errors_on_membership_cycle` test verbatim, under the existing exclusive coarse lock.
- Added an explicit depth cap `MAX_MEMBERSHIP_DEPTH = 16` (edges) enforced in `add_member_set`: rejects an edge whose resulting deepest chain (ancestors-of-parent + 1 + descendants-of-member) exceeds the cap, via two bounded recursive CTEs → new `MembershipDepthExceeded` error.

**Poster-lock cleanup (A2 gating) — rolled in** ([discussion](https://github.com/GaloyMoney/cala/pull/814#issuecomment-5205801542))
`find_for_update` now gates its SHARED membership-guard lock (classid `EC_SET_LOCK_CLASS`) on `NOT eventually_consistent`, exactly like the FOR_UPDATE lock beside it. Post-#811 the only EXCLUSIVE taker in that namespace is the `member_has_balance_history` attach-guard, and its fence is only complete on accounts where the poster writes balance rows *under* the lock — the **non-EC** accounts. On EC accounts the poster writes no balance rows (the streaming rollup does, in its own tx), so the SHARED lock there fenced nothing. Net: **posting into an EC-rooted subtree takes zero advisory locks on the EC nodes** (the discriminant is `eventually_consistent`, not leaf-vs-ancestor, so #813's EC plain leaves are covered too). `find_ec_balances_for_update` (rollup applier) is unchanged — shared/shared never conflicted.

*Recorded caveat (validated against the guard rule before landing):* the guard's `cala_entries` EXISTS could always race an **in-flight uncommitted first** posting to an EC member, because entries are inserted before the poster's lock prelude runs — the window existed pre-gating (insert→lock) and is now the full poster tx. Same race class, slightly wider; committed activity is still always seen (MVCC). Properly closed by the planned route-by-membership-seq change in the rollup applier; acceptable now since EC-only postings are rare until lana flips its omnibus accounts. Documented in the `find_for_update` and `balance/mod.rs` doc-comments.

New test `ec_poster_locks.rs::ec_only_posting_holds_no_membership_guard_locks`: `pg_locks` inspection on an isolated database — a synchronous-accounts posting holds ≥2 classid-1 SHARED locks mid-transaction (control) and zero after commit; an EC-only posting holds **zero** classid-1 locks mid-transaction.

**Not in this PR:** Locked-account handling.

## Tests
- `account_set.rs::errors_on_collision` kept **verbatim** — every original double-membership rejection (duplicate edge, account-in-child, account-in-sibling) passes unchanged against the explicit validation.
- New `account_set.rs::errors_on_double_membership`: diamond via shared ancestor, the same conflict inside one batch, set-level diamond while still empty, descendant-path overlap with all sets empty (`A⊃B`, `B⊃D`, `X⊃D` → `A.add(X)` rejected — bugbot finding), and subtree-account overlap on `add_member_set`.
- `account_set.rs`: `errors_on_membership_cycle` (from #800, one adjusted assertion — see above) + `errors_on_membership_depth_exceeded`.
- `account_set_membership_locks.rs`: the original protocol tests are kept (different-member adds run concurrently; same-member adds serialize; structure ops fence account-member ops) plus a new structure-ops-serialize test; the closure-materialization test is replaced by a walk-resolves-all-ancestors + no-closure-rows assertion.
- Existing balance / account_set / `ec_streaming_rollup.rs` tests exercise both resolvers indirectly and pass unchanged.
- Regenerated `.sqlx` (all `transitive` queries removed).

## Downstream ripple (action before ship)
Dropping the `transitive` column breaks any external SQL/tooling that reads it — **lana's own `.sqlx` and any ops dashboards / ad-hoc queries must be audited** before this ships. Membership *semantics* are unchanged: double membership remains prohibited exactly as before (same `MemberAlreadyAdded` error).

## Coordination
Builds on **#813** (ec-plain-accounts, now merged). The two are independent in `balance/repo.rs`: #813 added a *separate* `fetch_ec_leaf_accounts` (keyed on `is_account_set`) rather than extending `fetch_ec_set_mappings`, so the only shared surface was the baseline migration (drop `transitive` + #813's `is_account_set`, cleanly merged) and the derived `.sqlx` set (fully regenerated against `main`'s schema). See the poster-lock note above for the one place #813's guard changes interact with the original PR-3 plan.

---

> [!NOTE]
> **High Risk**
> Core ledger membership and balance rollup paths change materially (closure removal, explicit graph validation, EC lock gating); breaking schema for existing DBs and any SQL that referenced `transitive`.
> 
> **Overview**
> Removes **materialized transitive membership** — account→set membership is stored as **direct edges only**, and ancestor sets are resolved at read time via bounded recursive walks over set→set edges.
> 
> **Schema:** Drops the `transitive` column from `cala_account_set_member_accounts` in the baseline migration (fresh deploy; no in-place `ALTER`).
> 
> **Reads:** `fetch_mappings_in_op` and `fetch_ec_set_mappings` use `WITH RECURSIVE` upward walks; member listing queries no longer filter `transitive IS FALSE`.
> 
> **Writes:** Add/remove account or set members insert or delete a single direct edge; closure propagation CTEs are removed.
> 
> **Validation (replaces closure unique constraint):** `assert_no_double_membership` blocks second paths to the same set (`MemberAlreadyAdded`); `add_member_set` adds cycle detection (`MembershipCycleExceeded`), set/account path conflicts, and **`MAX_MEMBERSHIP_DEPTH` (16)**. Empty set-level double edges are rejected up front (tighter than closure-era behavior).
> 
> **Posting locks:** `find_for_update` skips classid-1 advisory locks when `eventually_consistent` is true, so EC-only postings hold zero membership-guard locks; non-EC behavior unchanged.
> 
> **Tests:** Cycle, depth, double-membership, lock protocol, walk-without-closure-rows, and `ec_poster_locks` coverage; `.sqlx` regenerated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 424ef71fc96db965c0ee828013c8bbfd88053fed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cala-ledger/src/account_set/repo.rs
```diff
@@ -644,22 +693,167 @@ impl AccountSetRepo {
...
+          ) AS "conflict!"
+          "#,
+            account_set_id as AccountSetId,
+            member_account_set_id as AccountSetId,
+        )
+        .fetch_one(db.as_executor())
+        .await?;
+        if set_conflict.conflict {
+            return Err(AccountSetError::MemberAlreadyAdded);
+        }
```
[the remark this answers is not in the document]
bodymindarts: Fixed in 5349465c — confirmed real. The set-level check now walks up from the member's **entire subtree** (member + descendants) instead of only the member's ancestors, so a descendant that already reaches the target chain through an edge bypassing the member (`A ⊃ B`, `B ⊃ D`, `X ⊃ D`, then `A.add(X)`) is rejected up front even while every set involved is empty. The subtree-walk subsumes the previous member-ancestors check, and the cycle check that runs first guarantees the subtree itself cannot legitimately intersect the target chain (no new false positives). Covered by a new all-empty descendant-path case in `errors_on_double_membership`.

cala-ledger/src/account_set/repo.rs
```diff
@@ -73,6 +81,69 @@ async fn lock_for_account_member_op(
...
+/// the account's existing containments; any set the same account would
+/// reach twice — via an existing path, via the new edge, or across two
+/// pairs of the same batch — is a violation. Surfaced as
+/// [`AccountSetError::MemberAlreadyAdded`], the same error the old
+/// closure's unique-constraint collision produced.
+///
+/// Must run under the account-member lock protocol: the walk is a read
+/// over the set graph and the account's direct edges, and the insert
+/// that follows relies on its result staying valid until commit.
+async fn assert_no_double_membership(
```
bodymindarts: remove as top level fn - put it on the repo... call via self.

cala-ledger/src/account_set/repo.rs
```diff
@@ -581,22 +623,13 @@ impl AccountSetRepo {
         account_id: AccountId,
     ) -> Result<(), AccountSetError> {
         lock_for_account_member_op(db, account_id).await?;
+        // Delete the single direct edge; there are no materialized
```
bodymindarts, line 630: line above - lock_for_account_member_op - again.. move all Fns here onto the structure.

cala-ledger/src/account_set/repo.rs
```diff
@@ -644,22 +693,179 @@ impl AccountSetRepo {
...
+        if account_conflict.conflict {
+            return Err(AccountSetError::MemberAlreadyAdded);
+        }
+
+        // Depth cap: the deepest root->leaf chain through the new edge is
+        // (edges above the parent) + 1 (this edge) + (edges below the
+        // member). Reject if that would exceed MAX_MEMBERSHIP_DEPTH so the
+        // read-time ancestor walk stays cheap and bounded. Both walks use
+        // UNION (dedup) and stop at the cap, so the check is itself O(cap).
+        let depth = sqlx::query!(
```
bodymindarts: can the 3 queries - depth cap and transitive membership check be unified somehow? Less round trips?

cala-ledger/src/account_set/repo.rs
bodymindarts: Done in c36d8bf9 — both `assert_no_double_membership` and `lock_for_account_member_op` are now private methods on `AccountSetRepo`, called via `self.`.

cala-ledger/src/account_set/repo.rs
bodymindarts: Done in c36d8bf9 — `lock_for_account_member_op` moved onto `AccountSetRepo` along with `assert_no_double_membership`; no free functions remain in the file (only the lock/depth consts).

cala-ledger/src/account_set/repo.rs
bodymindarts: Done in c36d8bf9 — all four validation queries (cycle check, set-level path uniqueness, account-level overlap, depth cap) are now one statement: the recursive walks are shared CTEs (`target_ancestors` doubles as cycle source + depth numerator + target-chain seed; `member_subtree` doubles as reach seed + depth denominator + account-overlap scope) and a single row returns `cycle` / `set_conflict` / `account_conflict` / `depth`. One round trip instead of four, same verdict precedence. Side benefit: the #800 cycle walk is now depth-capped, so it terminates even on a corrupted graph (complete under the enforced cap).

bodymindarts:
**Rolling one more small change into this PR** (from the perf-review pipeline — the last piece of the post-#811 lock cleanup): **stop taking the poster's SHARED advisory lock on eventually-consistent accounts.**

### Why it's dead weight now
After #811 deleted the watermark recalc, the only remaining EXCLUSIVE taker in the class-1 (`EC_SET_LOCK_CLASS`) namespace is the `member_has_balance_history` attach-guard (exclusive on the member). That guard only gets a complete fence from posters on accounts where the poster itself writes balance rows *under the lock* — i.e. **non-EC** accounts. For EC accounts the poster writes no balance rows (the streaming rollup does, in its own tx), so the SHARED lock `find_for_update` takes on every EC node fences nothing — pure per-posting overhead (one advisory lock per EC ancestor/leaf per `(account, currency)`).

### The change
In `BalanceRepo::find_for_update`'s lock statement, gate the shared lock exactly like the exclusive one beside it:

```sql
SELECT
    CASE WHEN NOT a.eventually_consistent THEN
        pg_advisory_xact_lock_shared($1::int4, hashtext(v.account_id::text))
    END,
    CASE WHEN NOT a.eventually_consistent THEN
        pg_advisory_xact_lock(hashtext(concat($2::text, v.account_id::text, v.currency)))
    END
FROM UNNEST($3::uuid[], $4::text[]) AS v(account_id, currency)
JOIN cala_accounts a ON a.id = v.account_id
ORDER BY v.account_id, v.currency
```

The discriminant is `eventually_consistent`, **not** leaf-vs-ancestor — with #813's EC plain accounts this correctly covers EC leaves too. Net effect: posting into an EC-rooted subtree takes **zero** advisory locks on the EC nodes.

Also:
- Update the doc-comment above the query ("SHARED lock … taken on *every* row — leaves and ancestors, EC and non-EC alike") to the new rule.
- `find_ec_balances_for_update` (rollup applier) takes SHARED on EC set-accounts; shared/shared never conflicted with posters anyway — no change needed there.

### One caveat, for the record
The attach-guard's `cala_entries` EXISTS has a fence gap on EC members *either way*: a poster inserts entries before it reaches the lock prelude, so an in-flight **first** posting to an EC account can commit concurrently with an attach that saw no entries (pre-join activity then folds into the parent via the walk). Today the SHARED lock narrows that window to entry-insert → lock-acquisition; after this change the window is the full poster tx. Same race class, slightly wider — properly closed only by the planned route-by-membership-seq change in the rollup applier (next pipeline step). Landing this now seems fine since the window already exists unfenced pre-lock, and EC-only postings are rare until lana flips its omnibus accounts.

### Test
A `pg_locks`-inspection test (pattern exists in `tests/account_set_membership_locks.rs`) asserting a posting that touches only EC nodes holds no classid-1 advisory locks; existing suite otherwise unchanged.

bodymindarts:
Landed in 424ef71f, after validating against every business rule established on this PR:

1. **Path uniqueness / no double membership** — enforced under the membership-graph locks (coarse `123456` + per-member class-2), a namespace posters never touch. Class-1 gating cannot interact. ✅
2. **MemberHasBalanceHistory guard** — non-EC members: SHARED retained on every non-EC row, and the poster writes the balance rows the guard checks *under* it → fence complete, unchanged. EC members: all **committed** activity is still always seen (MVCC — the EXISTS needs no lock for that). The only exposure is the in-flight uncommitted *first* posting you called out; I verified the linchpin in `post_transaction_in_op`: `entries.create_all_in_op` runs **before** `update_balances_in_op`, so the insert→lock window predated this change. Widening it to insert→commit is the same race class, per your caveat. Recorded in the `find_for_update` + `balance/mod.rs` doc-comments with a pointer to the route-by-membership-seq follow-up. ✅
3. **#802 FK (no entries to set-backing accounts)** — structural, lock-independent. ✅
4. **Cycle + depth cap** — coarse-EXCLUSIVE membership validation, class-1 irrelevant. ✅
5. **Streaming rollup sole EC writer / outbox ordering** — poster writes no EC balance rows either way; applier ordering doesn't depend on poster locks; `find_ec_balances_for_update` untouched (shared/shared never conflicted). ✅
6. **Lock-ordering doctrine (#684/#779/#810)** — the gated SHARED stays inside the same `ORDER BY`'d statement; the `CASE WHEN` wrapper matches the empirically-verified pattern of the exclusive lock beside it; removing acquisitions can't create new deadlock cycles. ✅
7. **Non-EC posting semantics** (read-your-write, A1 serialization) — both locks still taken on every non-EC row. ✅

Change is exactly your SQL. Test: `ec_poster_locks.rs::ec_only_posting_holds_no_membership_guard_locks` — `pg_locks` inspection on an isolated database; the synchronous-accounts control holds ≥2 classid-1 SHARED locks mid-tx and zero after commit, the EC-only posting holds **zero** classid-1 locks mid-tx. Full suite 120/120, `nix flake check` green, `.sqlx` regenerated. PR body updated (the 'deliberately NOT in this PR' section is replaced by the rolled-in change + recorded caveat).

PR #817, opened 2026-08-06
feat(balance): entry-sourced EC balance verify/repair tool (PR 5)

bodymindarts:
## Summary

Since the streaming EC rollup replaced `recalculate_balances_*` (#811), a drifted eventually-consistent balance had **no repair path**: nothing could recompute it. This adds the deterministic, bounded reconciliation tool from the perf-review roadmap (§6, PR 5):

- **`Balances::verify_ec(journal_id, &[AccountId])`** — read-only: recompute expected EC balances (set-backing accounts *and* EC plain leaves) from contributing entries and report per-layer/direction drift (`EcDriftReport`).
- **`Balances::repair_ec(...)`** — verify + append **corrective snapshots** (history stays append-only; nothing is rewritten or deleted) and rebuild the cumulative-effective series for effective-drifted pairs.

Implements the impl spec in drua library `cala-dev/impl-pr-ec-rebuild-tool.md`; master context is the holistic lock/perf review (`cala-dev/analysis-locks-holistic-perf-review.md`, memory `019fd1f5`).

## Correctness core

- **Consistency anchor.** The applier is delta-based and not idempotent, and outbox seq order ≠ commit order — so the *only* clean fold boundary is the rollup job's committed cursor `C` (obix delivery is gapless: everything ≤ C is committed **and** folded; everything > C folds later exactly once). Expected state = fold of entries whose transaction's `TransactionCreated` event has `seq ≤ C`.
- **Freezing `C`.** The reconciler takes **EXCLUSIVE** class-1 advisory locks on its targets (join-free UNNEST, Rust-sorted — canonical ordering) and reads `C` only *after*. The applier's flush takes the SHARED counterpart on every EC account it writes (`find_ec_balances_for_update`) in the same tx that advances the checkpoint — this PR is what that "defensive serialization point" was defending for. A target-touching batch blocks before writing *or* advancing `C`; its `(C, C']` window is disjoint from the `≤ C` fold; it applies on top of the corrected state after commit. No marker column, no job-pause API.
- **Routing = current membership walk** (downward inverse of #814's upward walk). Sound because the #813 membership guard freezes every node on an applied entry's ancestor path (leaf has entries; every ancestor — sync or EC — has history), so no edge on the path can be attached or cut.
- **Ordering/replay.** Entries fold via the existing `Snapshots::new_snapshot`/`update_snapshot` with `time = tx.created_at`, in `(tx event seq, entry sequence)` order — field-exact replay of applier semantics (values, versions, entry ids, timestamps). Entry *sets* come from `cala_entries` + `initialized` events (a checkpoint-straddling event group still counts in full); scans are keyset-chunked and memory-bounded (one running snapshot per pair — the #740 lesson).

## Deliberate calls (flagging per review)

- **Cursor accessor**: read-only `SELECT execution_state_json FROM job_executions WHERE job_type = 'cala.ec_balance_rollup'` (spec §6 open question — chose raw scoped SQL; can move to a job-crate API later).
- **Corrective outbox events**: settled snapshots have **no outbox event type at all** (only `EffectiveBalance*` exist), so there is nothing to publish for settled corrections. Effective reinserts publish through the normal `insert_new_snapshots` path — note the first reinserted row re-emits `EffectiveBalanceCreated` at `all_time_version = 1`.
- **Effective series rebuilt as one row per effective date** (`version` = entries that day, `all_time_version` = entries ≤ day). Intra-day per-entry rows and rewrite-inflated `all_time_version` counts are not reconstructible from the ledger (they depend on arrival-order interleaving); all read-path semantics (`effective ≤ date` lookups, `all_time_version` diffs as entry counts) are preserved.
- **Version drift is evidence, not a trigger**: `EcDriftReport` carries `expected_version`/`found_version`, but drift = value deltas — a corrective append legitimately leaves found = expected + 1.
- **Fresh-chain repair** (row + history wiped): a single corrective row carries the applier-identical end state; intermediate versions are not re-materialized.
- **RETENTION DEPENDENCY**: the fold requires the persistent outbox to cover full history. obix range-partitions (#809) but never prunes today. If outbox pruning/archival ever lands, this needs an alternative "entries ≤ C" mapping — documented in the module docs.

## Coordination with PR 6 (attach fence — sibling, in flight)

- All reconcile SQL lives in a **new** `balance/reconcile/` module. The only `balance/repo.rs` contact is a one-line `EC_SET_LOCK_CLASS` visibility change; `find_for_update`, `find_ec_balances_for_update`, the B3 member guards and the class-1 doctrine comment are untouched (PR 6 owns that surface).
- PR 6 must keep the applier's SHARED lock in `find_ec_balances_for_update` — it is this tool's serialization partner (both specs carry the note; whoever lands second re-checks the doctrine comment).
- Until PR 6 lands, posters also hold SHARED on EC ancestors, so a repair chunk briefly blocks posting into the target subtree; after PR 6 only the applier serializes against repair.

## Tests (`tests/ec_reconcile.rs`, isolated DB per test)

- clean verify → zero drift; repair is a no-op
- corrupted `latest_values` → **exact** drift deltas reported → repair → second verify clean; corrective at `version + 1`
- current + history wiped → rebuilt from entries (fresh chain, applier-identical version)
- **cursor = 0 double-count guard**: posted-but-unapplied entries contribute nothing (a naive Σ-entries "repair" would be double-counted by the applier later); stream then catches up exactly once
- exclusive lock blocks the applier before write/checkpoint (observed via `pg_locks`), then exactly-once catch-up after release
- back-dated effective series: interior date dropped + latest corrupted → effective drift flagged with settled clean → whole series rebuilt
- streaming continues on top of the corrected state after repair

`nix flake check` green (fmt / clippy `-D warnings` / deny / audit), full live-PG suite 128/128, `.sqlx` regenerated.

## Semver

Additive (**minor**): new public methods + `EcDriftReport`/`LayerDrift` + `BalanceError::NotEventuallyConsistent`. No schema changes, no migrations.

## Downstream

- **lana**: should expose this as an admin runbook / mutation (`verify_ec` sweep + gated `repair_ec`) — flagged only, not touched here.
- **obix**: no changes; retention dependency flagged above.

### DIFF
--- a/cala-ledger/.sqlx/query-186980c34d237000a9f94f822fd68ee79e6b891843678dd7a93d1aab87f00b42.json
+++ b/cala-ledger/.sqlx/query-186980c34d237000a9f94f822fd68ee79e6b891843678dd7a93d1aab87f00b42.json
@@ -1,0 +1,22 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n        SELECT execution_state_json\n        FROM job_executions\n        WHERE job_type = $1\n        ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "execution_state_json",
+        "type_info": "Jsonb"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text"
+      ]
+    },
+    "nullable": [
+      true
+    ]
+  },
+  "hash": "186980c34d237000a9f94f822fd68ee79e6b891843678dd7a93d1aab87f00b42"
+}
--- a/cala-ledger/.sqlx/query-20f3a51b4e92e80fe18b5e82fb1892b8f7520c87de4f2fdfa1613c00994b9e70.json
+++ b/cala-ledger/.sqlx/query-20f3a51b4e92e80fe18b5e82fb1892b8f7520c87de4f2fdfa1613c00994b9e70.json
@@ -1,0 +1,28 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n        WITH RECURSIVE subtree AS (\n            SELECT v.target_id AS target_id, v.target_id AS set_id\n            FROM UNNEST($1::uuid[]) AS v(target_id)\n            UNION\n            SELECT s.target_id, e.member_account_set_id\n            FROM subtree s\n            JOIN cala_account_set_member_account_sets e\n                ON e.account_set_id = s.set_id\n        )\n        SELECT\n            s.target_id AS \"target_id!: AccountId\",\n            m.member_account_id AS \"leaf_id!: AccountId\"\n        FROM subtree s\n        JOIN cala_account_set_member_accounts m\n            ON m.account_set_id = s.set_id\n        ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "target_id!: AccountId",
+        "type_info": "Uuid"
+      },
+      {
+        "ordinal": 1,
+        "name": "leaf_id!: AccountId",
+        "type_info": "Uuid"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "UuidArray"
+      ]
+    },
+    "nullable": [
+      null,
+      false
+    ]
+  },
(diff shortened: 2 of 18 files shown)


bodymindarts:
closing for now until we have a concrete need.

PR #824, opened 2026-08-08
test: add cargo-fuzz workspace + fix fuzz-found panics in cel-interpreter

nicolasburtey:
## Summary

Adds cargo-fuzz (libFuzzer) infrastructure for the library's in-memory attack surface, plus fixes for three panic paths the fuzzer found in the first ~10 minutes of running.

### Fuzzing workspace (`fuzz/`)

Three targets, committed seed corpora, and a README with run instructions:

| Target | Covers |
| --- | --- |
| `cel_compile` | CEL compilation of arbitrary strings (upstream ANTLR parser, big-stack compile thread, compilation cache) |
| `cel_evaluate` | Compile + evaluate against a production-like context (`params`, velocity `entry`/`time`, layer constants) + all `TryFrom<CelResult>` coercions |
| `core_types_json` | serde round-trips of `ParamDefinition`, `VelocityLimitValues`, `TxTemplateValues`, `EntryValues`, `Layer`, `Currency` |

Run with `cargo fuzz run -s none <target>` (`-s none` because the default ASan build needs nightly; debug assertions stay on).

### Bugs found & fixed here

- **`timestamp.format('%Q')` panic** — invalid chrono format specifiers make `DelayedFormat`'s `Display` impl return `fmt::Error`, which panics inside `.to_string()`. Now returns an `ExecutionError`. Any velocity control / tx template with a bad format string could previously crash the server.
- **Bytes→JSON coercion panic** — `TryFrom<CelResult> for serde_json::Value` hit `unimplemented!()` on `CelValue::Bytes` (e.g. `{'a': b'x'}`). Now a `ResultCoercionError`.
- **Upstream `cel` parser panic made non-fatal** — the parser's `RecursionListener` depth counter underflows on adversarial input. `compile_program` now converts compile-thread panics into `CelParseError` instead of crashing the caller.

All three have regression tests.

### Reported upstream (not fixed here)

- [cel-rust#305](https://github.com/cel-rust/cel-rust/issues/305) — the depth-counter underflow panic (fix suggested, verified locally)
- [cel-rust#306](https://github.com/cel-rust/cel-rust/issues/306) — **exponential memory blowup**: a ~125-byte invalid expression consumes >1 GiB during `Program::compile` (×5 repeats → >6 GiB). This is a remote DoS against any deployment compiling user-supplied CEL; no in-process mitigation exists. Needs an upstream fix or a sandboxing decision on our side.

## Test plan

- [x] `cargo test -p cala-cel-interpreter` — 13 tests pass, incl. 3 new regression tests
- [x] `cargo clippy -p cala-cel-interpreter -p cala-ledger-core-types` — clean
- [x] ~2M fuzz executions across all three targets after the fixes — no further crashes
- [ ] CI


nicolasburtey:
Superseded by two split PRs: #825 (cel-interpreter fixes) and #826 (fuzz infrastructure).

PR #841, opened 2026-08-11
refactor(account-set): isolate graph validation

vindard:
## Summary

The parent PR keeps set-membership validation beside the existing account-membership validator to match Cala’s current layering. That leaves pure graph invariants coupled to cache state and makes algorithmic behavior harder to test without infrastructure.

This follow-up moves both account and set path decisions into a dedicated graph-validation module. The graph cache now only adapts snapshots and overlays, selects SQL fallback when data is incomplete, and delegates deterministic decisions to the functional core. Behavior and lock ordering are unchanged.

### Chain

- Parent: [#840](https://github.com/GaloyMoney/cala/pull/840)

## Test plan

- [x] `cargo test -p cala-ledger`
- [x] `cargo clippy -p cala-ledger --tests -- -D warnings`
- [x] Unit-test account path acceptance, duplicate ancestry, and unknown-set fallback
- [x] Unit-test set cycles, duplicate paths, account-path conflicts, and depth attribution
- [x] Independently compare behavior against the parent branch

---

> [!NOTE]
> **Medium Risk**
> Membership graph validation is correctness-critical; this is a structural move with delegated logic and new unit tests, but any mistake could change reject/accept behavior for attaches.
> 
> **Overview**
> This PR **extracts** account-set membership graph rules out of `graph_cache.rs` into a new **`graph_validation`** module so the cache only adapts snapshots/overlays and picks SQL fallback when the graph view is incomplete.
> 
> **Account attach path:** `assert_no_double_membership_in_op` now passes snapshot/overlay lookups into **`has_duplicate_account_membership_paths`** (replacing the removed in-cache path counter). **`None`** still means unknown set → SQL walk; **`Some(true)`** → `MemberAlreadyAdded`.
> 
> **Set attach path:** **`validate_set_memberships`** holds the combined-graph checks (cycles, duplicate paths, account reachability, depth cap). `assert_valid_set_memberships_in_op` only fetches repo data and delegates there.
> 
> **`MAX_MEMBERSHIP_DEPTH`** moves to `graph_validation`; `repo.rs` imports it from there for SQL depth checks. Module docs in `graph_cache` note the new layering.
> 
> **Tests** in `graph_validation` cover account path cases and set validation errors without cache/DB infrastructure. Intended behavior and lock ordering are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 41e847d31cd74d606a8d78c4f8497486310bc7a4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cala-ledger/src/account_set/graph_cache.rs
+++ b/cala-ledger/src/account_set/graph_cache.rs
@@ -19,9 +19,9 @@
 //!
 //! Layering: `AccountSets` (service) calls this cache; this cache holds
 //! a handle on `AccountSetRepo` and orchestrates — ALL SQL lives in
-//! `repo.rs`, cache state and account-path graph logic live here, and pure
-//! set-membership invariants live in `graph_validation.rs`. The call graph is
-//! strictly service -> cache -> repo; the repo never calls back up into the cache.
+//! `repo.rs`, cache state and snapshot adaptation live here, and pure graph
+//! invariants live in `graph_validation.rs`. The call graph is strictly
+//! service -> cache -> repo; the repo never calls back up into the cache.
 //!
 //! Why the split read is safe — the two inputs have opposite write
 //! rates:
@@ -92,7 +92,7 @@ use crate::primitives::{AccountId, AccountSetId, JournalId};
 
 use super::{
     error::AccountSetError,
-    graph_validation::validate_set_memberships,
+    graph_validation::{has_duplicate_account_membership_paths, validate_set_memberships},
     repo::{AccountSetRepo, SetGraphNode},
 };
 
@@ -464,12 +464,31 @@ impl SetGraphCache {
             Some(Overlay { parents, meta })
         };
 
-        match Self::count_membership_paths(
-            &snapshot,
-            overlay.as_ref(),
+        let is_known = |set_id: &AccountSetId| -> bool {
+            snapshot.meta.contains_key(set_id)
+                || overlay
+                    .as_ref()
+                    .is_some_and(|overlay| overlay.meta.contains_key(set_id))
+        };
+        let parents_of = |set_id: &AccountSetId| -> &[AccountSetId] {
+            snapshot
+                .parents
+                .get(set_id)
+                .or_else(|| {
+                    overlay
+                        .as_ref()
+                        .and_then(|overlay| overlay.parents.get(set_id))
+                })
+                .map(Vec::as_slice)
+                .unwrap_or(&[])
+        };
+
+        match has_duplicate_account_membership_paths(
             account_set_ids,
             account_ids,
             &probe.seeds,
+            is_known,
+            parents_of,
         ) {
             Some(false) => {
                 span.record(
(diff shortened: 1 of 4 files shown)


bodymindarts (commented):
Reviewed as a delta against `56110589` (#840 head). **This is a clean, behavior-preserving extraction and it's the better-shaped of the two PRs.** I verified the move rather than trusting it:

- `validate_set_memberships`, `graph_has_path`, `first_depth_overflow`, `graph_max_depth` — bodies are byte-identical to their `graph_cache.rs` originals; the only change is dropping the `Self::` qualifier. Pure move. ✅
- `count_membership_paths` → `has_duplicate_account_membership_paths` — the `known`/`parents_of` closures were hoisted from callee to caller. Semantics preserved exactly, **including the `.or_else()` precedence** (snapshot first; overlay consulted only when the snapshot has no entry for that set). `Option<&Overlay>` → `Option<Overlay>` + `.as_ref()` is equivalent. Seeding still preserves pair multiplicity, `None`-on-unknown and `Some(true)`-on-count>1 unchanged. ✅
- `repo.rs:12` now imports `MAX_MEMBERSHIP_DEPTH` from `graph_validation` — **this resolves the layering finding** (r3754996733) properly: `graph_validation` is a leaf with no dependency on `repo` or `graph_cache`, so both import downward. This is the right fix, better than leaving the const in `repo.rs`. ✅
- Visibility is correct — `mod graph_validation` is private in `account_set/mod.rs`, and `pub(super)` scopes the items to the `account_set` module where both consumers live. No stale `graph_cache::MAX_MEMBERSHIP_DEPTH` references remain. ✅
- CI green (Bugbot pass, performance-report pass).

**The unit tests close two of the three coverage gaps I raised on #840**, which is the most valuable thing here:

- `set_paths_reject_a_duplicate_existing_and_proposed_path` is exactly the "conflict through a committed edge" case — existing `[(root,branch),(branch,leaf)]` + proposed `[(root,leaf)]`. In #840 *every* rejection test started from an empty graph, so the existing-edge interaction had zero coverage.
- `set_paths_reject_an_account_reachable_twice` covers the `account_paths` branch, which was entirely untested in #840 — the branch whose failure mode is silent balance double-counting.
- `set_paths_accept_a_valid_combined_tree` also exercises existing+proposed together on the accept side.

## Is it mergeable as-is?

**No — but for stacking reasons, not code reasons.**

`baseRefName` is `perf/batch-set-memberships`, and GitHub reports `mergeStateStatus: BLOCKED`. It cannot merge until #840 does, and I've requested changes on #840 over the validation read scope.

There's also a concrete **rework risk** in landing this extraction before #840's data-loading question is settled: if #840 takes the suggestion to source edges from `GraphSnapshot.parents` under the epoch guard, `validate_set_memberships` most likely stops taking `&[(AccountSetId, AccountSetId)]` slices and starts taking accessor closures — the same shape `has_duplicate_account_membership_paths` already uses. That would churn the signature this PR just froze, plus its tests.

So my recommendation stands from the #840 review: **fold these two commits into #840** rather than keeping the stack. You'd settle the layering fix, the unit-test gap, and the read-scope design in one reviewable unit, instead of extracting a pure core and then immediately reshaping its interface. If you'd rather keep them separate, then settle #840's data-loading shape first and rebase this on top.

## Nits (all minor, none blocking on their own)

### 1. The termination proof was dropped in the move

`count_membership_paths`'s doc comment is gone, and with it the load-bearing paragraph:

> Terminates unconditionally: every pop increments some count and the second increment of any count returns immediately, so pops are bounded by (known sets + 1) per account. A corrupted cyclic graph revisits a set and reports a conflict — a conservative reject, where the SQL walk's unbounded UNION ALL recursion would not terminate at all.

Also gone: the note that this counts **paths**, not reachability (unlike `expand`, which dedups via a visited set). That's precisely the subtlety a future editor is most likely to "optimize" away by adding a visited set — which would silently break the check. The replacement doc (`graph_validation.rs:13-15`) is three lines and keeps neither. For a PR whose stated purpose is making the pure core legible and testable, losing the correctness argument is a regression — please carry both paragraphs across.

### 2. `validate_set_memberships` still has no doc comment

`graph_validation.rs:49`. It was undocumented in #840 too, but there it sat directly under a documented caller. It's now the module's primary entry point and the four verdicts it enforces (cycle → set-path uniqueness → account-path uniqueness → depth, in that precedence order) are only discoverable by reading the body. The precedence order in particular is a deliberate semantic inherited from `add_member_set` (`repo.rs:708-731`) and worth stating.

### 3. `proposed_edges[0]` is now a sharper edge

`graph_validation.rs:101` panics if `proposed_edges` is empty *and* the cycle branch is taken. Reachable only with a corrupted (cyclic) committed graph plus an empty batch — but note `set_paths_reject_an_account_reachable_twice` already calls this with `proposed_edges = &[]`, so empty input is now a live, tested shape at the module boundary, and the only thing keeping it safe is the empty-check in `mod.rs:364` two layers up. A `debug_assert!(!proposed_edges.is_empty())`, or just returning a cycle error keyed off the existing edges, would pin it locally.

### 4. Unit-test cases still worth adding while you're here

The harness makes these ~4 lines each:
- self-edge `(A, A)` → `MembershipCycleDetected` (distinct code path from the multi-node cycle: caught via `indegree`, then attributed by the `account_set_id == member_account_set_id` arm rather than `graph_has_path`)
- the same proposed edge twice, and a proposed edge duplicating an existing one → `MemberAlreadyAdded`
- depth overflow **caused by combination with existing edges** — `set_paths_attribute_the_first_depth_overflow` passes `existing = &[]`, so `first_depth_overflow`'s handling of a non-empty existing graph is untested, and that's the part where the binary search's prefix semantics actually interact with prior state

---

To be clear about where this leaves the stack: the extraction itself I'd approve on sight. Holding off only because it's gated on #840 and its interface may move with #840's resolution.

bodymindarts:
Approving to merge into the other PR

bodymindarts (approved):

vindard:
Folded into #840 so the pure graph validators and their unit tests land with the batch API. The two commits were cherry-picked as bc7e7df1 and b2dee501.

PR #843, opened 2026-08-11
perf(account-set): steady-state small-batch improvements

vindard:
This PR assumes `add_member_sets` will also be used for steady-state small batches, not only the one-shot chart import. It adds the remaining optimizations and lock-protocol coverage Justin flagged for that shape.

## Changes

- `perf(account-set): fast path for single-edge batches` — route a one-edge batch through the existing single-attach CTE path instead of combined-graph validation, preserving the old cost model for degenerate batches.
- `docs(account-set): note single-edge fast path in cost model` — document the degenerate-case behavior in the public API doc.
- `test(account-set): batch membership lock protocol coverage` — prove that batch structure mutations serialize with single-edge structure mutations and fence account-member mutations on the same exclusive coarse lock; also verify a single-edge batch produces the same persisted state and outbox event as a direct set attach.

## Test plan

- [x] `cargo test -p cala-ledger`
- [x] `cargo clippy -p cala-ledger --tests -- -D warnings`
- [x] New lock-protocol tests for batch-vs-single serialization and batch-vs-account-member fencing
- [x] New single-edge batch fast-path test

This is stacked on #840.

---

> [!NOTE]
> **Medium Risk**
> Touches account-set hierarchy mutation routing and the exclusive membership lock protocol, but the change is a small early-return that reuses the existing single-attach path.
> 
> **Overview**
> Optimizes `add_member_sets_in_op` for degenerate one-edge batches by routing them through the existing single-attach CTE path instead of combined-graph validation and cache work.
> 
> Also documents this cost-model behavior and adds lock-protocol tests proving batch structure mutations still serialize with single-edge structure ops and fence account-member mutations, plus a test that a single-edge batch persists the same edge and outbox event as a direct attach.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 516e8f259a638a093a2cd7e9fec9dd68dcfde096. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cala-ledger/src/account_set/mod.rs
+++ b/cala-ledger/src/account_set/mod.rs
@@ -430,6 +430,18 @@ impl AccountSets {
             });
         }
 
+        // A single edge is cheaper through the existing single-attach path:
+        // one targeted CTE validation and insert, no combined-graph cache
+        // work. This preserves the old cost model for callers that batch a
+        // single pair by accident or for convenience.
+        if members.len() == 1 {
+            let (account_set_id, member_account_set_id) = members[0];
+            return self
+                .repo
+                .add_member_set(op, account_set_id, member_account_set_id)
+                .await;
+        }
+
         self.repo.lock_for_set_membership_op(op).await?;
         self.set_graph_cache
             .assert_valid_set_memberships_in_op(op, members)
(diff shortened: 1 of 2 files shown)


bodymindarts:
seems superfluous...

PR #845, opened 2026-08-13
perf(posting): consolidate the write path into lock/read/write + batch API

bodymindarts:
## What

Restructures `post_transaction` into three fused statements and adds a batch
posting API. Every hot-path statement takes arrays, so posting one transaction
is the N=1 case of posting a batch.

**15.05 → 6.05 client round trips per posting.** Exact, no variance,
machine-independent — measured with `pg_stat_statements`.

| # | phase | statement |
|---|-------|-----------|
| 1 | | `BEGIN` |
| 2 | lock (the fence) | union advisory locks + `now()` + template versions |
| 3 | read | memberships + graph epoch + journals + account meta + velocity controls + balances |
| 4 | write | transactions + entries + both event streams + balance snapshots |
| 5 | | outbox insert (obix commit hook) |
| 6 | | `COMMIT` |

FK integrity checks are unchanged at 9.00/posting (server-side, no round trip).

## Why the three phases can't collapse further

Both boundaries are correctness requirements, not style:

- **lock → read**: an advisory-lock wait *inside* a statement does not refresh
  that statement's snapshot. Any read the locks protect — balance rows, and
  above all account-set memberships — must run in a *later* statement, or a
  concurrent attach committing while we block would be invisible. This is the
  stale-read class the attach fence exists to close.
- **read → write**: snapshot versioning and velocity CEL enforcement run in
  Rust between them. Nothing is written until every posting has been folded and
  enforced, which is what lets a rejection name the offending posting with no
  rows to undo.

## Batch API

`CalaLedger::post_transactions{,_in_op}` taking `Vec<PostingInput>`.
`post_transaction` delegates as N=1. All-or-nothing; mixed journals and
templates allowed in one batch.

Ordering is exactly as if the postings ran one at a time, in input order, inside
one transaction: later postings observe earlier ones' balances, velocity
enforces against the chained snapshots, snapshot versions increment in order.

**Concurrent batches are deadlock-free by construction** — one canonically
sorted union lock batch over every posting's entry pairs, then one sorted
ancestor batch, over disjoint key classes. The old per-posting loop took locks
across posting boundaries with no global ordering, which is what made
unpartitioned batching unsafe before.

A batch is bounded by **distinct accounts, not size** — the fence holds two
advisory locks per distinct entry account in Postgres' *shared* lock table.
Past `MAX_DISTINCT_BALANCES_PER_BATCH` the flow refuses up front rather than
letting Postgres raise a bare `out of shared memory`, which names neither cause
nor fix and can strike unrelated concurrent transactions.

## Correctness

- 151/151 tests pass, including new `cala-ledger/tests/transaction_batch.rs`
  (12 tests): chained balances on overlapping accounts, batch-equals-loop
  equivalence, mixed journals/templates, all-or-nothing rollback with index
  attribution, locked-account and duplicate-id rejection, concurrent
  overlapping batches, effective-balance equivalence across mixed effective
  dates (full row sets compared including `all_time_version`), velocity
  equivalence across a batch, and the distinct-account ceiling.
- The doctrine tests that encode the locking design — `attach_posting_fence`,
  `ec_poster_locks`, `lock_accounts`, `set_graph_cache`, `velocity`,
  `effective_balance`, `balance`, `account_set` — pass **unmodified**. They were
  the regression net for the restructure.
- `nix flake check` green: clippy `-D warnings`, fmt, deny, audit.
- Ledger invariants after ~1.5M postings: settled debits equal credits per
  journal, imbalance exactly 0; 200/200 sampled accounts match the sum of their
  entries.

## Behaviour changes worth reviewing

- **`LedgerError` gained `PostingError`.** Posting-path failures now surface as
  `LedgerError::PostingError(...)` rather than the flat variants — e.g. velocity
  enforcement is `PostingError::VelocityError(VelocityError::Enforcement(_))`.
  `PostingError::Rejected { index, tx_id, reason }` attributes a batch failure
  to one posting. Downstream `match` sites need updating.
- **Locked accounts are now rejected regardless of `eventually_consistent`.**
  Previously the status check was skipped for EC accounts because the poster
  doesn't write their balances; locked is a business fact, not a
  balance-maintenance detail.
- **Timestamps unchanged.** `NOW()` is `transaction_timestamp()`, pinned at
  `BEGIN`, so returning it from the lock statement yields the identical value
  the dedicated `SELECT NOW()` did. Manual clocks still win via
  `COALESCE($manual, NOW())`.

## Notes

- Measured with a local throughput/statement-budget harness that is
  deliberately **not** in this PR — this is the `cala-ledger` change only.
- Follow-up, not included: `update_effective_balances` groups by
  `(journal, effective date)` at 2 statements per group, so a batch spanning
  many back-dates loses amortization on that portion. It works around
  `SnapshotOrEntry::cmp` declaring same-date `(Entry, Snapshot)` unreachable;
  ordering `Entry` after `Snapshot` would allow a single pass. That change lands
  in `balance/effective`, so it is out of scope here.
- Follow-up, not included: the lenient/partial-acceptance batch API. The phase
  layout makes it nearly free — every business-level rejection now surfaces
  client-side *before* the write statement.

---

> [!NOTE]
> **High Risk**
> Rewrites the core posting path—balance locks, snapshots, velocity, and account-set fan-out—in one fused transaction; regressions would affect financial correctness or concurrency safety despite extensive tests.
> 
> **Overview**
> **Consolidates the ledger posting hot path** into a new `posting` module. `post_transaction` / `post_transaction_in_op` now delegate to `Postings::post_all_in_op` with a one-element batch, so single and batch posts share the same phased flow and array-oriented SQL.
> 
> The flow is **lock → read → in-memory fold/enforce → write → outbox**: one statement for advisory locks, `now()`, and template version probes; one fused read (set-graph epoch/seeds, journals, accounts, velocity controls, balances); one fused insert for transactions, entries, events, and balance snapshots. Transactions and entries are no longer created via their entity repos’ persist hooks—`PostingRepo` writes rows and the posting flow publishes outbox events.
> 
> Adds **`post_transactions` / `post_transactions_in_op`** (`Vec<PostingInput>`), all-or-nothing with `PostingError::Rejected { index, tx_id, reason }` for attributable pre-write failures, plus **`BatchTooManyAccounts`** when distinct locked balances exceed the shared advisory-lock budget.
> 
> **Set-graph resolution** now consumes the membership probe from the posting read (no separate probe round trip); **velocity** enforces per batch with controls already loaded and skips DB work when no limit windows match.
> 
> **API/error changes:** failures surface as `LedgerError::PostingError(...)`; locked accounts are rejected even when `eventually_consistent`. Template hot-path resolution moves to `TemplateCache` with version re-check on the fence statement.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7b6d9916e765f669bde02344dd5f7859987f4102. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bodymindarts:
Both confirmed and fixed — thanks, the first one was a real bug.

**Cross-journal ancestor mappings (High) — confirmed, fixed in d570be4f.**

I reproduced it before fixing. A leaf account carries no `journal_id` of its
own, so the same leaf can be a member of sets in several journals.
`resolve_ancestors` resolved per journal *correctly*, then flattened the results
into one `leaf -> sets` map — and `Snapshots::from_entries` takes the set id
from that map but the `journal_id` from the **entry**. Failing test, before the
fix:

```
set B under journal A must have no balance row, got Some(1290)
```

So journal B's set acquired a full balance row under journal A. Worse than
wrong data: that row was never locked, because `ancestor_keys` is built inside
the per-journal loop from the *unflattened* result, so the spurious pair isn't
in the lock batch at all.

Velocity had the identical hazard — `VelocityBalanceKey` takes `journal_id`
from the entry and `account_id` from the mappings, so a limit attached to a
foreign journal's set would be enforced, and its balance written, under this
journal.

This was introduced by the batching: the pre-consolidation path called
`fetch_mappings_in_op` once per posting with that posting's own journal, so the
results never merged. Single-journal batches were never affected.

Fix: mappings are now `journal -> (leaf -> sets)` (`AncestorMappings`), and each
consumer indexes by the journal it is already iterating — `fold_balances`,
`update_effective_balances`, and velocity via each posting's
`transaction.journal_id`. Regression test:
`a_multi_journal_batch_does_not_cross_ancestor_sets_between_journals`, which
posts one transaction per journal in a single batch with the same leaves
attached to a set in each, and asserts both that each set gets its own journal's
balance and that neither gets a row in the other's.

**Inner posting span is INFO (Low) — fixed in 7b6d9916.**

Correct, and it contradicted the module's own convention: every other span in
`posting/` was already `level = "debug"`. Demoted, matching 46f8290d (#778).

153 tests pass, `nix flake check` green.

PR #867, opened 2026-08-27
ci: replace Cachix with local Nix caches

krtk6160:
## Summary

- vendor GaloyMoney/galoy-concourse-shared#26
- use persistent /nix-cache volumes for check-code, integration-test, and fuzz
- remove the Cachix token and cache-name pipeline values

## Dependency

Land GaloyMoney/galoy-concourse-shared#26 first so the vendored commit is on main.

## Validation

- vendir sync completed against the pinned shared commit
- ytt renders the full Cala pipeline successfully
- rendered pipeline contains no Cachix credential or command references
- vendored shell scripts pass syntax checks


krtk6160:
Superseded by #869, which stacks the non-vendor local Nix cache changes onto the Concourse-generated shared-files bump.

PR #887, opened 2026-09-18
chore(deps)!: bump es-entity to 0.13.0, job to 0.15.2, obix to 0.12.1

bodymindarts:
## Summary
- Bumps `es-entity` 0.12.21 → 0.13.0, `job` 0.14.0 → 0.15.2, `obix` 0.11.0 → 0.12.1 — all breaking 0.x bumps.
- **es-entity 0.13.0**: `PaginatedQueryRet` gates construction behind `::new(entities, has_next_page, end_cursor, requested_size)` and the `entities` field behind `.entities()`/`.into_parts()` (#229). Every repo that hand-built the struct now calls `::new(...)` with the query's `first` as `requested_size`; callers that read `.entities` move to `.entities()` or `.into_parts()`.
- **obix 0.12.0**: replaces the commit-ordered envelope with a lane typed into the delivery itself (#156). `SingletonSubscriber::handle_persistent` now takes `&EventDelivery<P, L>` instead of `&Arc<PersistentOutboxEvent<P>>` (retention via `event.inner().clone()`), `Subscription::await_sequence` → `await_position`, and `SubscriptionError::CaughtUpTimeout`/`SubscriptionStreamStatus` report the dynamic `StreamPosition` instead of a lane-typed sequence. cala only runs the default insert lane, so `EcRollupStatus` and `LedgerError::EcCaughtUpTimeout` are updated to match with no behavior change.
- **obix 0.12.0** also replaces the commit-log partition tables with a sparse `persistent_outbox_commit_checkpoints` table, and **job 0.15.0** adds `job_waiters` plus `job_executions.woken_at` for wake-on-completion. Both migrations are updated in place — regenerated verbatim from the upstream templates (byte-for-byte round-trip verified after stripping the `cala_` prefix) — since nothing is live on this schema yet.
- **job 0.15.2** / **obix 0.12.1** are trailing patch releases that only strip comments from their migration templates; the cala-prefixed copies here are updated to match, with no schema change.
- `.sqlx` query cache regenerated against the new schema.

## Test plan
- [x] `SQLX_OFFLINE=true cargo build --locked --workspace --all-targets --all-features`
- [x] `SQLX_OFFLINE=true cargo clippy --workspace --all-targets --all-features` (clean)
- [x] `cargo nextest run --workspace --all-features` against a fresh local Postgres migrated from these migration files — 219/219 passed
- [x] `cargo run -p cala-ledger --bin event-schemas --features json-schema` — no schema drift
- [x] `cargo doc --no-deps` — no new warnings

---

> [!NOTE]
> **High Risk**
> Rewrites job and persistent-outbox migrations and job scheduling SQL while changing pagination and outbox delivery APIs—high impact on background jobs, EC rollup lag/wait semantics, and any consumer of paginated list types.
> 
> **Overview**
> **Breaking dependency upgrade** for `es-entity` 0.13.0, `job` 0.15.2, and `obix` 0.12.1, with matching `Cargo.lock` and regenerated `cala-ledger/.sqlx` offline query metadata.
> 
> **es-entity:** Pagination results are no longer built with struct literals. Repos construct `PaginatedQueryRet::new(..., query.first)` and callers read pages via `.entities()`, `.into_parts()`, or updated pagination continuation patterns in tests.
> 
> **obix:** Persistent delivery moves to `EventDelivery` in `handle_persistent` (EC rollup clones via `event.inner()`). Subscription waiting uses `await_position` and `StreamPosition` in errors/status instead of lane-typed sequences. The outbox commit path drops partitioned `cala_persistent_outbox_commit_log` + state tables for sparse `cala_persistent_outbox_commit_checkpoints` (migrations rewritten in place).
> 
> **job:** Schema adds `job_waiters`, `job_executions.woken_at`, and updated queue promote/reschedule/retry SQL (wake-on-completion semantics). Job setup migration is trimmed/regenerated to match upstream templates.
> 
> **Ledger surface:** `EcRollupStatus` is built from explicit `checkpoint()` / `frontier()`; `LedgerError::EcCaughtUpTimeout` carries `StreamPosition`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8f358991a6cb7193f25969ef66ecb209b9840298. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cala-ledger/src/entry/mod.rs
```diff
@@ -134,7 +134,8 @@ impl Entries {
                 Default::default(),
             )
             .await?
-            .entities;
+            .into_parts()
+            .0;
```
bodymindarts, line 138: ugly - I thought `into_parts()` was internal only... why is it pub?

cala-ledger/tests/balance.rs
```diff
@@ -32,21 +32,23 @@ fn all_balances_query<C: std::fmt::Debug>() -> es_entity::PaginatedQueryArgs<C>
...
-        .entities
+        .into_parts()
+        .0
         .into_iter()
         .map(|balance| (balance.details.currency, balance))
         .collect()
 }
 
-fn balances_by_id<C>(
+fn balances_by_id<C: std::fmt::Debug>(
```
bodymindarts, line 46: Ugly - why does `C` now carry this constraint? We shouldn't mandate that constraint on library users.

cala-ledger/src/entry/mod.rs
vindard: I think this is an anti-pattern. I left `into_parts` because I'm thinking want _both_ entities and the next query if we need the entities. Having `entities` alone is a signal that we are ignoring future pages. The intention is to fetch `let (entities, next) = query.into_parts()` or `let next = query.into_next_query()` but never `let entities = query.into_entities()` and never this tuple destructuring style, since this might signal a bug if we're ignoring next pages?

bodymindarts: okay - can you PR here to show what should be idiomatic?

PR #889, opened 2026-09-21
refactor(pagination)!: migrate to es-entity 0.14.0

woodjd12:
Upgrade to es-entity 0.14.0 and migrate pagination consumers to its `Page`
API. Existing traversal loops retain their behavior; callers that consumed
one page explicitly discard its continuation. Existing pagination bugs are
deferred to separate PRs.

This crosses public Rust type boundaries and requires a breaking release.

Dependencies and rollout:

- Job: https://github.com/GaloyMoney/job/pull/223
- Obix: https://github.com/GaloyMoney/obix/pull/161
- Review and release in order: Job → Obix → Cala → Lana.
- Job and Obix are pinned to their draft commits so this branch can be
  checked out independently. Replace those pins with published versions and
  refresh the root and fuzz lockfiles before merging/releasing.

Validation:

- Workspace check with all targets and features.
- All 14 existing transaction integration tests passed after the final
  behavior-preserving revision.
- 73 ledger unit tests and 43 account-set, balance, effective-balance, and
  entry-filter integration tests passed during migration validation.
- Fuzz targets compiled; formatting and diff checks passed.
- One existing `unused_mut` warning in `transaction_batch.rs` remains.


nicolasburtey:
Superseded by #892: same es-entity 0.14 pagination migration, but against the published crates.io releases (es-entity 0.14.1, job 0.16.1, obix 0.13.1) instead of git-rev forks. Released as cala 0.31.0.

